PR #20, opened 2015-04-01
Extended interface to the InMemoryReadModelStore

tabro:
Added some methods to the InMemoryReadModelStore:

- GetAll(): does what it says
- Find(Func<TReadModel, bool> predicate): Returns all readmodels that matches the provided predicate.


rasmus:
@tabro Nice :smile:

PR #31, opened 2015-04-13
Generate SQL for creating read model tables

### LINKED ISSUE #22: Help with creation of SQL for MSSQL read models
Instead of developers writing their SQL manually for MSSQL read models, create a helper classes that can assist them.

Its vital that this is opt-in.

rasmus:
Fixes #22


rasmus:
Don't do this, especially after the MSSQL read model cleanup

PR #45, opened 2015-04-25
Create ITimeMachine

### LINKED ISSUE #43: Introduce ITimeMachine
Create a service for getting `DateTimeOffset.Now` to make it easier to test time.

rasmus:
Create a `ITimeMachine` that developers can provide their own implementation of if they wish. Fixes #43


rasmus:
We don't wanna add dependencies for the aggregate, so this is a bad solution. Wait until someone actually has a good idea (and need) before implementing this

PR #52, opened 2015-04-30
Aggregate ID

rasmus:


rasmus:
@MartinHave What do you think about this solution?

MartinHave:
@rasmus Nice one.

* We should consider the name ```IAggregateId``` What about entities and such? Whould ```IIdentity``` be a better suited name?

* Why do we want to limit the input to the identity with the name prefix? Is it really necessary? I understand the idea with descriptive ids. Can we simplify?

rasmus:
@MartinHave The provided implementation `AggregateId` is merely a generic ID that is "good enough" for most, you can provide another implementation if you want as it merely needs to implement `IAggregateId`.

rasmus:
@MartinHave I considered `IIdentity` but my OCD couldn't handle the double `I`. But if you have a good argument why `IIdentity` is better, then I'm OK with that.

PR #55, opened 2015-05-02
Apply missing events to read models

### LINKED ISSUE #41: Apply missing events to read models
If a read model version doesn't match with the domain events to be applied, missing events should be loaded and applied to the read model.

rasmus:
Fixes #41


rasmus:
Find some other way of doing this

PR #60, opened 2015-05-11
Read model improvements

### LINKED ISSUE #58: Read models that span multiple aggregates
The current implementation of read models only take events from a single aggregate. We need a method for creating read models than span multiple aggregates.

Problems

- As optimistic concurrency within one aggregate is handled by the event store, how to handle when events are applied to a read model that spans multiple aggregates and the events are committed simultaneously.

rasmus:
When done it fixes #58


rasmus:
@MartinHave If you have the time, have a look at this

MartinHave:
@rasmus I will give it a look this afternoon :)

MartinHave:
@rasmus I like it. Nice and clean.

PR #64, opened 2015-05-19
Populate read models from event store

### LINKED ISSUE #63: Build all read models from event store
We need support for flushing all read models and then building them again from the event store. Its critical that both the purge and rebuild is controlled explicitly by the developer.

rasmus:
Fixes #63 when done


rasmus:
Closed this depends on a global sequence number, which is to be removed

PR #70, opened 2015-05-22
Read model populate

### LINKED ISSUE #63: Build all read models from event store

rasmus:
Fixes #63 when done


rasmus:
@MartinHave I have made it possible to recreate read models based on all events. There are some problems that really bug me, but lets discuss those.

PR #89, opened 2015-08-20
Command IDs

rasmus:
Command IDs can be used for a few different things.

- Doing truly asynchronous execution of commands using e.g. RabbitMQ. Commands are basically value objects, so it would be possible to send them using RabbitMQ. The command IDs will allow applications that send the commands, to receive a notification when the command was executed and with a success code.
- Traceability and auditing as the command ID can be logged


rasmus:
@MartinHave Could you give some feedback on this feature?

JC008:
Can the deterministic GUID factory be used here in conjunction with a identity composer? 

The command members may be composed into hash that is passed into the GUID factory, thereby uniquely identifying incoming commands and allowing for dropping duplicates/repeats.

rasmus:
@JC008 The use of members to create IDs seems a bit too specific. However, the ID will be possible to set by developers, making it possible to do this in their own implementation.

JC008:
Specific is the goal - member values == unique identity. If the [`TComposer`](https://github.com/rasmus/EventFlow/pull/102#issuecomment-134820766) concept is used it does not have to be a default behavior.

rasmus:
@JC008 Good point regarding the `TComposer` and (as you stated) I don't think it should be the default behavior as some domains might allow multiple commands with the same arguments.

rasmus:
Article related to the PR.

http://blog.sapiensworks.com/post/2015/08/26/How-To-Ensure-Idempotency/

rasmus:
@JC008 I been reading up on your suggestion for creating comman IDs based on member values and will add support for it in the core package. It will not be the default behavior, but it will be easy to enable (hopefully).

rasmus:
@JC008 What do you think of the `DeterministicIdCommand` I added here? You requested something like this.

MartinHave:
@rasmus I like this feature. Also the ```DeterministicIdCommand``` could come in handy at some point. I agree that it should not be the default option.

:+1:

Source/EventFlow/Aggregates/IAggregateRoot.cs
```diff
@@ -35,8 +35,13 @@ public interface IAggregateRoot
         IEnumerable<IAggregateEvent> UncommittedEvents { get; }
         bool IsNew { get; }
 
-        Task<IReadOnlyCollection<IDomainEvent>> CommitAsync(IEventStore eventStore, CancellationToken cancellationToken);
+        Task<IReadOnlyCollection<IDomainEvent>> CommitAsync(
+            IEventStore eventStore,
+            ISourceId sourceId,
```
MartinHave: @rasmus This is a breaking change. Consider extending the IAggregateRoot interface with the old implementation. If the aggregate does not care about the sourceId then it can be omitted.

rasmus: @MartinHave I know. The change is to a "internal" service not typically used by developers, but I added the change to the release note along with a description on how to get a new ID

JC008:
@rasmus The solution will work very well - may I suggest calling it `DistinctCommand`.

rasmus:
@JC008 Better name, I have renamed it and will write some documentation.

PR #95, opened 2015-08-23
Enable use of the "resolver" aggregate root factory in the core package

rasmus:
@JC008 I merged your PR and thought of a few additions that would be nice, e.g. be able to use the resolver factory from the core package.

The core and Autofac packages currently share a few files.


rasmus:
@JC008 You can review the minor changes I did here

PR #98, opened 2015-08-24
WIP: Configuration alternative

### LINKED ISSUE #36: Alternative interface for configuration
We need to rethink how the `EventFlowOptions` interface is and how users define configuration.

Proposals:

``` csharp
  // Option 1
  EventFlowOptions.New
    .As.EventSource().Use<MsSqlEventSource>()

  // Option 2
  EventFlowOptions.New
    .EventStore(r => r.Use<MsSqlEventStore())
```

@tabro Any other suggestions?

rasmus:
Fixes #36, see issue for discussion.

### DIFF
--- a/Source/EventFlow.Tests/IntegrationTests/EventStores/FilesEventStoreTests.cs
+++ b/Source/EventFlow.Tests/IntegrationTests/EventStores/FilesEventStoreTests.cs
@@ -53,7 +53,7 @@ namespace EventFlow.Tests.IntegrationTests.EventStores
 
                 var resolver = eventFlowOptions
                     .UseInMemoryReadStoreFor<InMemoryTestAggregateReadModel>()
-                    .UseFilesEventStore(FilesEventStoreConfiguration.Create(storePath))
+                    .EventStore.UseFilesEventStore(FilesEventStoreConfiguration.Create(storePath))
                     .CreateResolver();
 
                 _configuration = resolver.Resolve<IFilesEventStoreConfiguration>();
--- a/Source/EventFlow/Configuration/IConfigureEventStore.cs
+++ b/Source/EventFlow/Configuration/IConfigureEventStore.cs
@@ -1,0 +1,29 @@
+﻿// The MIT License (MIT)
+//
+// Copyright (c) 2015 Rasmus Mikkelsen
+// https://github.com/rasmus/EventFlow
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software is furnished to do so,
+// subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+namespace EventFlow.Configuration
+{
+    public interface IConfigureEventStore
+    {
+        EventFlowOptions Options { get; }
+    }
+}
--- a/Source/EventFlow/EventFlow.csproj
+++ b/Source/EventFlow/EventFlow.csproj
@@ -70,6 +70,7 @@
     <Compile Include="Commands\CommandHandler.cs" />
     <Compile Include="Commands\ICommandHandler.cs" />
     <Compile Include="Configuration\EventFlowConfiguration.cs" />
+    <Compile Include="Configuration\IConfigureEventStore.cs" />
     <Compile Include="Configuration\IEventFlowConfiguration.cs" />
     <Compile Include="Configuration\IResolverContext.cs" />
     <Compile Include="Configuration\IServiceRegistration.cs" />
@@ -113,6 +114,7 @@
     <Compile Include="Exceptions\NoCommandHandlersException.cs" />
     <Compile Include="Exceptions\OptimisticConcurrencyException.cs" />
     <Compile Include="Exceptions\WrongImplementationException.cs" />
+    <Compile Include="Extensions\ConfigureEventStoreExtensions.cs" />
     <Compile Include="Extensions\DateTimeOffsetExtensions.cs" />
     <Compile Include="Extensions\DictionaryExtensions.cs" />
     <Compile Include="Extensions\DisposableExtensions.cs" />
--- a/Source/EventFlow/EventFlowOptions.cs
+++ b/Source/EventFlow/EventFlowOptions.cs
@@ -39,11 +39,20 @@ namespace EventFlow
 {
     public class EventFlowOptions
     {
+        private class ConfigureEventStore : IConfigureEventStore
+        {
+            public EventFlowOptions Options { get; }
+            public ConfigureEventStore(EventFlowOptions options) { Options = options; }
+        }
+
+
         public static EventFlowOptions New => new EventFlowOptions();
 
         private readonly ConcurrentBag<Type> _aggregateEventTypes = new ConcurrentBag<Type>();
         private readonly EventFlowConfiguration _eventFlowConfiguration = new EventFlowConfiguration();
         private Lazy<IServiceRegistration> _lazyRegistrationFactory = new Lazy<IServiceRegistration>(() => new AutofacServiceRegistration()); 
+
+        public IConfigureEventStore EventStore => new ConfigureEventStore(this);
 
         private EventFlowOptions() { }
 
--- a/Source/EventFlow/Extensions/ConfigureEventStoreExtensions.cs
+++ b/Source/EventFlow/Extensions/ConfigureEventStoreExtensions.cs
@@ -1,0 +1,43 @@
+﻿// The MIT License (MIT)
+//
+// Copyright (c) 2015 Rasmus Mikkelsen
+// https://github.com/rasmus/EventFlow
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software is furnished to do so,
+// subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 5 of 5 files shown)


JC008:
@rasmus I find from using EventFlow that the configurability is key to my adoption of EventFlow as it smooths you into the concepts in EventFlow. So, a simpler base line configuration should not be overly complicated while retaining the ability to create detailed configurations as and where needed.

I don't full comprehend the suggested solution (possibly a sub module configuration strategy ?) and also see an added level of indirection.

Perhaps a standard approach and guidance plus module style composition will suffice.

rasmus:
@JC008 I agree, but I wanted to get started on it and have a look a different possibilities. How you configure a third-party lib and what you can customize is something that's I think is one of the most important aspects of creating a lib, especially if used in a large enterprise application. That and a license similar to the MIT license.

Its not something that I think will be added anytime soon as I really want to make sure that it right. And backwards compatibility is also a concern.

Alignment of what the configuration options definitely the first step.

JC008:
How about a generic singleton like `ModuleRegistration<>`. You could add more specialized `[Type]Registration<>` classes to wrap the registration of special cases. This could also allow for reporting on the current configuration and opportunities to do detailed configuration in an `IModule` or detect conflicting configurations and resolving them using e.g. an `IConfigurationResolver` etc. 

```c#
class Program
{
    static void Main(string[] args)
    {
        var configuration = new Configuration();
        configuration.Register(SqlEventStore.Module);
        configuration.Register(SqlReadModels.Module);
        configuration.Register(Books.Module);
        configuration.Register(Magazines.Module);

        // retrieve a module 
        var configFromType = configuration.Module<SqlEventStore>();
        var configFromInterface = configuration.Module<IEventStoreConfiguration>();

        // retrieve a value
        var sqlConnectionString = configuration.Module<SqlEventStore>().ConnectionString;
    }
}

public interface IModule
{
    void Register(EventFlowOptions eventFlowOptions);
}

public interface IEventStoreConfiguration
{
	string ConnectionString { get; }
}

public class Configuration
{
    public Configuration()
    {
        EventFlowOptions = EventFlowOptions.New;
    }

    public void Register(IModule module)
    {
        Modules.Add(module);
    }

    public T Module<T>()
    {
        // search for T
        throw new KeyNotFoundException();
    }

    public EventFlowOptions EventFlowOptions { get; }

    protected IList<IModule> Modules { get; } = new List<IModule>();
}

// a generic singleton that can be made lazy and allow for using interfaces and retain a "static calling style"
abstract public class ModuleRegistration<T>
    where T : IModule, new()
{
    public static readonly T Module;

    static ModuleRegistration()
    {
        Module = new T();
    }

    protected ModuleRegistration()
    {
        if (Module != null)
        {
            throw new InvalidOperationException("use '.Module' to retrieve the module");
        }
    }
}

public class Books : ModuleRegistration<Books>, IModule
{
    public void Register(EventFlowOptions eventFlowOptions)
    {
        // call methods on eventFlowOptions
        throw new NotImplementedException();
    }
}

public class Magazines : ModuleRegistration<Magazines>, IModule
{
    public void Register(EventFlowOptions eventFlowOptions)
    {
        // call methods on eventFlowOptions
        throw new NotImplementedException();
    }
}

public class SqlEventStore : ModuleRegistration<SqlEventStore>, IModule, IEventStoreConfiguration
{
    public string ConnectionString { get; }

    public void Register(EventFlowOptions eventFlowOptions)
    {
        eventFlowOptions
            .ConfigureMsSql(MsSqlConfiguration.New
                .SetConnectionString(TestDatabase.ConnectionString))
            .UseEventStore<MsSqlEventStore>();
    }
}

public class SqlReadModels : ModuleRegistration<SqlReadModels>, IModule
{
    public void Register(EventFlowOptions eventFlowOptions)
    {
        eventFlowOptions
	        .ConfigureMsSql(MsSqlConfiguration.New
		        .SetConnectionString(TestDatabase.ConnectionString))
	        .UseMssqlReadModel<Books.BooksReadModel>();
    }
}


```

rasmus:
@JC008 I this its a good idea, but it could be simplified a bit leaving more flexibility for the developers.

Also, there are two parts, configuration and registration and I think they should be separate.

Here's another example of your code, but with a minor variation.

```csharp
public interface IEventFlowModule
{
  EventFlowOptions Register(EventFlowOptions eventFlowOptions);
}

// This in the extentions namespace
public static EventFlowOptions RegisterModule<T>(
  this EventFlowOptions eventFlowOptions)
  where T : new(), IEventFlowModule
{
  var eventFlowModule = new T();
  return eventFlowModule.Register(eventFlowOptions);
}

// Then define a module like this
public class SqlReadModels : IEventFlowModule
{
  public EventFlowOptionsRegister(EventFlowOptions eventFlowOptions)
  {
   return eventFlowOptions
     .ConfigureMsSql(MsSqlConfiguration.New
       .SetConnectionString(TestDatabase.ConnectionString))
     .UseMssqlReadModel<Books.BooksReadModel>();
    }
}

// And then register the module
var config = EventFlowOptions.New
  .RegisterModule<SqlReadModels>()
  ...
```

One of the main things I try to do with EventFlow is to ensure that you can run two separate instances of EventFlow in the same process. This means that the only things that put in `static` fields, is reflection caches as these would be the same for multiple instances.

I think this is an important point, as it also allows you do to testing in parallel.

Regarding configuration, here I have considered doing something I did for another project.

```csharp
// Define configuration
public interface IMsSqlConfiguration
{
  [ConfigurationKey("eventflow.mssql.connectionstring")]
  string ConnectionString { get; }
  
  [ConfigurationKey("eventflow.mssql.connectionstring")]
  [DefaultConfiguration(false)]
  bool RetryOnTimeoutException { get; }
}

// Read configuration
var configurationReader = resolver.Resolve<IConfigurationReader<IMsSqlConfiguration>>();
var connectionString = configurationReader.Get(c => c.ConnectionString);
```

Then allow developers to implement a store that meets their needs and that the `ConfigurationReader` uses. Here are a few examples of what I have done before.

- Application settings store, basically wraps `ConfigurationManager.AppSettings`
- An in-memory configuration store, which would be the default for EventFlow
- Settings specific for a user in a web application, using the current principal to get the user ID and then fetch the specific settings for that user

rasmus:
@JC008 I have thought about your suggestion a bit more and think your solution is good, I just needed to sleep on it  a bit.

I'll add it to EventFlow ASAP.

Your suggestion regarding putting the configuration alongside the code that uses it also makes a lot of sense.

JC008:
EventFlow can be used to create a true application server (e.g. not just a message bus behind a web server) but with the understanding that there is no single "server" and therefor application "services" are provided with the resulting requirement for a configuration system to match.

The DDD element in EventFlow naturally supports micro-services based on aggregates where `IAggregateRoot<>` represents transaction boundaries which creates the opportunity for a distributed application model. Coupled with the CQRS element of `ICommand<>` and `ICommandHandler<>` with the `IReadModelFor<>` as well as `IQuery<>` interfaces, EventFlow provides the foundation to build large scale distributed solutions.

While taking the above into consideration, it is important to remember that EventFlow is one of the many elements in an application which require configuration and may not own all the configuration data. For example, in a desktop application it could be used behind WPF while in a web application it needs to be configured alongside Asp.Net. Other projected variations requires using EventFlow with a Rpc configuration (JsonRpc) behind NanoMsg and queues/jobs using RabbitMq or Azure Queues are essential. Nonetheless, the registration and configuration in EventFlow applications will become substantive and be critical to it's usage and perceived reliability as well the stability of EventFlow and its implementations.

As an example, currently EventFlow requires the use of `AutofacServiceRegistration.CreateResolver(...)` fronted by `EventFlowOptions.CreateResolver(...)` to *kick start* EventFlow. In an application a call to `Autofac.ContainerBuilder.Build(...)` from an application level *bootstrapper* will not result in EventFlow being configured even though the extension call to `UseAutofacContainerBuilder(...)` appears to provide an inversion of control. EventFlow rightly uses the shared external `ContainerBuilder` for the registration of its services but then takes control of the `ContainerBuilder` by requiring `CreateResolver(...)` be the last method called in the container configuration process, thereby making the implementation if an application bootstrapper directly dependent on `EventFlowOptions`. The use of an `Autofac.Module` will mitigate this dependency inversion and should be considered.

The above example should be considered in refactoring the configuration system while keeping the overall requirements stated for EventFlow to wit, providing an open architecture with reasonable defaults and opportunities for specialized enhancements.

#### Composition (registration)

The following may be considered:

-   the [reactive manifesto](http://www.reactivemanifesto.org/)
-   registration and **re-configuration** can be reactive  
-   support for dynamic micro services
-   dynamic application composition based on various factors
    -   time of month
    -   time of day (e.g. catching up on projections)
    -   load on servers
-   evolving generational architecture and application domains
    -   route to Application Domain based on command/vesion/region etc
    -   reroute after reconfigure on failures (IRetryStrategy implementations)
    -   distributed and partitioned for use in both server and desktop solutions
    -   decoupled internal application dependancies are possible (iow micro-services)

#### Configuration

The following may be considered:

-   support for [12 factor](http://12factor.net/) style applications
-   integrations to [ZooKeeper](https://zookeeper.apache.org/), [Consul](https://consul.io/), [Etcd](https://coreos.com/etcd/), ConfigurationManager.AppSettings, Json files
-   IDE support when composing a module's configuration
    -   avoid lambdas as primary reader mechanism
    -   but leave `Func<T,I>` as usable extension for power users
    -   use namespaces or nested classes to manage hierarchy and enable further composition thereby avoid long configuration names
-   logging configurations to provide a review opportunity of historic operational configurations
-   interrogating relevant configuration information by configuration provider e.g. `if not has X then use ConnectionString from Z`
-   provide a `ConfigurationException` class  
-   support configuration validators - e.g. `IConfigurationValidator`
-   visibility e.g. see configuration in use on a status board (similar to mod\_status in Apache)
-   avoid duplicate meta information
    -   use public read/write members of class
    -   but allow for attribute based overrides
    -   avoid the [oddness](http://www.codeproject.com/Articles/16466/Unraveling-the-Mysteries-of-NET-2-0-Configuration) of the ConfigurationManager.AppSettings architecture, rather serialize objects to json
    -   use explicit configuration classes and interfaces
    -   single responsibility for configuration classes means avoiding the need to add extra "markup"

#### Multiple instances of EventFlow

This is an aspect I have not considered, but can see the need for it and it adds real value.

-   [Application Domain](https://msdn.microsoft.com/en-us/library/2bh4z9hs%28v=vs.110%29.aspx) support should be considered
-   Application Domains may be used to
    -   load and unload EventFlow service compositions
    -   provide hot configuration switchover
    -   support dynamic reconfiguration of services at runtime
    -   route commands to handlers based on predicates
-   [NUnit.ApplicationDomain](https://bitbucket.org/zastrowm/nunit.applicationdomain/><https://bitbucket.org/zastrowm/nunit.applicationdomain/) provides the ability to similate Application Domains in NUnit which is usefull for integration/behavioural testing

rasmus:
@JC008 Many thanks for the detailed description and I greatly value your input. I'll have to re-read your comment to make sure that I get all the details. I'll get back to it.

But of the right after reading the comment about the `CreateResolver()` in EventFlow, it bothers me as well. It really needs to be the last thing to do and if you don't want to create the resolver at that point, then your stuck. I have considered adding a `Initilize()` or `Build()` that does the initial configuration of EventFlow, i.e., parsing the added event types and adding the default implementations for the services that haven't yet been overloaded, but I haven't found a good way of providing the required functionality while keeping it simple and intuitive.

rasmus:
@JC008 I also think you are right in what you said about EventFlow being just a part of a larger application. EventFlow shouldn't try do dominate or influence how you would like to do e.g. configuration or configure your IoC container. Something I may have forgotten and need to keep in mind in the future.

JC008:
Autofac supports an [`IStartable`](http://docs.autofac.org/en/latest/lifetime/startup.html) interface that could solve the issue of being notified when the container is ready to run and allow EventFlow configure its service definitions.

JC008:
`IStartable` has potential to solve the ownership issue of the container build task. I did a test in https://github.com/JC008/AutofacPatterns/tree/master/AutoStart

rasmus:
Great, that would make it possible for the `CreateResolver()` to actually do nothing but what its name suggests and allow developers to omit calling it if they wanted to build the container somewhere else. Thanks for looking into that.

rasmus:
I created #133 and will look into it this week as its really something that has been bothering me. EventFlow should not dictate when you want to build/initialize your IoC container.

rasmus:
Much of this is implemented and the API for configuring EventFlow isn't going to change much at the moment.

PR #99, opened 2015-08-25
Added generic Identity validator support using IIdentityValidator and…

JC008:
… Identity<,>

- formalize Identity validation and create a reusable foundation using IIdentityValidator
- makes use of SingleValueObject and IIdentity
- Identity<,> matches behavior of Identity<>
- Identity<,> expects identifiers to be strings from an IO perspective in the same way as Identity<>

### DIFF
--- a/Source/EventFlow.Tests/EventFlow.Tests.csproj
+++ b/Source/EventFlow.Tests/EventFlow.Tests.csproj
@@ -76,6 +76,7 @@
     <Compile Include="IntegrationTests\InMemoryConfiguration.cs" />
     <Compile Include="IntegrationTests\ReadStores\InMemoryReadModelStoreTests.cs" />
     <Compile Include="IntegrationTests\ResolverTests.cs" />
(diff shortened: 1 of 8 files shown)


rasmus:
@JC008 Good idea to cleanup identity validation. I think you should rework the existing `Identity<>` so that it uses the new `Identity<,>` class and putting the current validation into e.g. `DefaultIndentityValidator`.

Regarding the ordering of static members, I think your suggestion of moving all initializers into static constructors is best. I'll go though the code this week and do just that.

After this is completed I'll move the identity related classes to the `EventFlow.Core` namespace, as an identity can be used for other entities that need an identity. Specifically commands and events.

Also, I'm considering using deterministic GUIDs for events using a combination of aggregate name, aggregate ID and the event aggregate sequence number (version) and to create the ID.

JC008:
Oi - I pushed to many files not related to the change yesterday - please close / discard - I will do a clean branch and pull request with a title matching your style.

PR #100, opened 2015-08-25
Can we use @event instead of e where possible?

JC008:
- or maybe domainEvent
- one character parameter names can be confusing


rasmus:
Yes, definitely. It will always be either `aggregateEvent` or `domainEvent`.

PR #102, opened 2015-08-25
Use deterministic GUIDs to create event IDs

rasmus:
Not sure if this is a good idea...

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,8 @@
    when name, i.e., the metadata property with key `aggregate_name` (or
    `MetadataKeys.AggregateName`). If you are dependent on the previous naming,
    use the new `AggregateName` attribute and apply it to your aggregates
+ * Breaking: Moved `Identity<>` and `IIdentity` from the `EventFlow.Aggregates`
+   namespace to `EventFlow.Core` as the identities are specific for aggregates
  * New: Aggregate names can now be configured using the attribute
    `AggregateName`. The name can be accessed using the new `IAggregateRoot.Name`
    property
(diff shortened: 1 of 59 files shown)


JC008:
This is a very good idea and will add great value to EventFlow. 

For example - I need to correlate aggregateroots based on a deterministic underlying element - an email address - to locate registration requests and map them to existing accounts and users. 

I am considering adding a `Identity<TIdentity, TComposer, TValidator>` abstract class where `TComposer` is used to compose the _payload_ of the identity and this will be one of the available composers.

A related issue I have is in the planned use of EventFlow in distributed solution of desktop, server, web and mobile clients where aggregates may be created anywhere but need to be deterministicely identifiable across the network for reconciliation purposes. The same goes for events in the use case mentioned e.g. `Who` `Changed` `What` `Where` `When` `Why` to track colliding edits to aggregates in an optimistic/late locking approach.

Another potential benefit is dropping the aggregate/class name  from the identity and making it a pure `GUID` as the aggregate object name is tracked in the meta data.

rasmus:
@JC008 Thanks for the feedback. The factory for creating the UUIDs will be added as a generic factory that merely returns a `Guid`, making it possible for everyone to use it. I'll add support for it in the base `Identity<>` class by adding a new `New(Guid namespaceId, string name)` making it easier to create aggregate root IDs if you are using that base class for creating the IDs.

rasmus:
@JC008 If you have a link to an article/book that discusses your use cases, I would like to read it.

JC008:
http://book.mixu.net/distsys/ (chpt 3 and 4)

Support for Lamport clocks is worth considering too.

rasmus:
@JC008 Thanks

PR #103, opened 2015-08-26
Make Identity composable and usable with all classes which require id…

JC008:
…entity support e.g. Aggregates, Events, Commands

- Note: Identity<,,>.New() is a method and could cause a lot of cascading changes
- see questions and notes

### DIFF
--- a/Source/EventFlow.Tests/UnitTests/Aggregates/AggregateIdWithValidatorTests.cs
+++ b/Source/EventFlow.Tests/UnitTests/Aggregates/AggregateIdWithValidatorTests.cs
@@ -1,0 +1,134 @@
+﻿// The MIT License (MIT)
+//
+// Copyright (c) 2015 Rasmus Mikkelsen
+// https://github.com/rasmus/EventFlow
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software is furnished to do so,
+// subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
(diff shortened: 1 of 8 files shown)


Source/EventFlow.Tests/UnitTests/Aggregates/AggregateIdWithValidatorTests.cs
```diff
@@ -0,0 +1,134 @@
...
+using EventFlow.Aggregates;
+using FluentAssertions;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace EventFlow.Tests.UnitTests.Aggregates
+{
+    internal class AggregateIdWithValidatorTests
```
rasmus: @JC008 These tests aren't executed as they are contained within a `internal` class. Look at the `CreatedIsSame()`, it should fail as the IDs doesn't contain a `-`

rasmus:
@JC008 After some thought I think this might be overthinking it a bit for a general purpose framework and that the generation of GUIDs in the format `[ID type]-[GUID]` is sufficient for most use cases. I agree with the points you made in #102 regarding deterministic GUIDs is good and I will add it. However, as EventFlow uses interfaces whenever possible, it's possible for you to use your own implementation of `IIdentity` in your application the meets your specific needs.

Does it make sense or am I just rambling?

JC008:
I was thinking the same - this is a little too complex for something so simple - a custom `IIdentity` implementation should suffice. I think its my reaction to being caught out by the GUID focus of `Identity` and lack of deterministic identity behavior by the default `Identity`. This will be solved by your solution for deterministic GUIDs.

Source/EventFlow.Tests/UnitTests/Aggregates/AggregateIdWithValidatorTests.cs
```diff
@@ -0,0 +1,134 @@
...
+using EventFlow.Aggregates;
+using FluentAssertions;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace EventFlow.Tests.UnitTests.Aggregates
+{
+    public class AggregateIdWithValidatorTests
```
rasmus: @JC008 Lacking the `[TestFixture]` attribute, the tests aren't executed

rasmus: @JC008 Hmmm... that's odd. Don't know why the test aren't executed

rasmus: @JC008 Sorry, they are... nevermind.

JC008:
@rasmus Your goal of keeping EventFlow focused yet extensible should guide on this. I suggest I add this code to the EventFlow Guide I am working on as an example of leveraging the architecture and design of EventFlow to achieve non EventFlow core goals - e.g. re-usable Id rules etc. It will achieve the goal of keeping EventFlow clean, focused and extensible while avoiding the inevitable library bloat and at the same time provide guidance on what is possible.

rasmus:
@JC008 Makes sense and good input as I really do want developers that use EventFlow to customize the parts that they don't want or can't use. All developers can't agree on everything, which is why I try to keep the default implementation very generic and leave the specifics others. If there are places that cannot be customized or that it would make sense to include another strategy in the default implementation and it does compromise a clean architecture, then please create an issue with the suggestion.

As soon as you get the EventFlow guide ready, let me know, and I'll add a link to the main read me file.

JC008:
@rasmus Agreed - e.g. AggregateState which was too deep to handle externally but added immense value.

PR #109, opened 2015-08-30
Guide on performance tips

rasmus:


rasmus:
Will try to do this by providing an alternative to the MSSQL event store schema. But not now.

PR #111, opened 2015-08-30
Larger example domain

rasmus:
Can be used for

- Example on how to unitize EventFlow
- Better scenario tests for read stores and event stores


rasmus:
Find another example, e.g. something with users as that might be closer to what developers would actually use EventFlow for

PR #113, opened 2015-09-04
Prevent abstract AggregateRoot<,> implementations from being registered as services

JC008:
…ed as services

- missing check for abstract registrations added
- AggregateExtensionsTests added to check behavior
- assembly helper added to EventFlow.Tests

QUESTIONS
- should the test  for abstracts not be moved to AutofacServiceRegistration.ValidateRegistrations()


rasmus:
@JC008 As soon as the build is fixed, I'll merge it. Remember to add the fix for the release notes.

JC008:
Help!

PR #115, opened 2015-09-04
Split event sourcing from aggregate roots

rasmus:
Instead of having event sourcing functionality directly implemented into aggregate roots, it should be put into a separate base class. This will allow "other stuff" to be event sourced.

Its _critical_ that there's no breaking change.


rasmus:
Bad idea, this introduces too many breaking changes at the moment

PR #116, opened 2015-09-05
Sagas

### LINKED ISSUE #17: Introduce a simple saga system
Inspiration
 - https://msdn.microsoft.com/en-us/library/jj591569.aspx
 - http://blog.jonathanoliver.com/cqrs-sagas-with-event-sourcing-part-i-of-ii/

Notes
 - Its critical that the saga system is optional, as developers might a different view on how to write sagas or even if they should be used

rasmus:
See if its possible to add saga support to EventFlow using the existing aggregate roots.

Fixes #17


Source/EventFlow.Tests/UnitTests/Sagas/SagaTestClasses.cs
```diff
@@ -0,0 +1,160 @@
...
+            ISagaHandles<SagaTestAggregate, SagaTestAggregateId, SagaTestEventB>,
+            ISagaHandles<SagaTestAggregate, SagaTestAggregateId, SagaTestEventC>
+        {
+            public TestSaga(TestSagaId id) : base(id)
+            {
+            }
+
+            public Task ProcessAsync(IDomainEvent<SagaTestAggregate, SagaTestAggregateId, SagaTestEventA> domainEvent, CancellationToken cancellationToken)
+            {
+                Publish((cb, c) => cb.PublishAsync(new SagaTestBCommand(domainEvent.AggregateIdentity), cancellationToken));
```
rasmus: @MartinHave As we discussed. What do you think of this concept, i.e., commands are published _after_ the aggregate is successfully stored to the event store? If I publish the command directly here from the aggregate, then the aggregate state will not have been updated (isn't stored yet) and the next part of the saga isn't done as the saga is only started by the `SagaTestEventA` event.

Source/EventFlow.Tests/UnitTests/Sagas/SagaTestClasses.cs
```diff
@@ -0,0 +1,160 @@
...
+            public Task ProcessAsync(IDomainEvent<SagaTestAggregate, SagaTestAggregateId, SagaTestEventA> domainEvent, CancellationToken cancellationToken)
+            {
+                Schedule(new SagaTestBCommand(domainEvent.AggregateIdentity));
+                Emit(new SagaEventA());
+                return Task.FromResult(0);
+            }
+
+            public Task ProcessAsync(IDomainEvent<SagaTestAggregate, SagaTestAggregateId, SagaTestEventB> domainEvent, CancellationToken cancellationToken)
+            {
+                Schedule(new SagaTestCCommand(domainEvent.AggregateIdentity));
```
rasmus: @MartinHave This was what I was talking about. The problem:

- A saga reacts to a event
- The saga updates and publishes a command (order is irrelevant)
- Command results in another event being published
- As the command bus is synchronous, the saga isn't committed to the event store yet and cannot react correctly to the newly emitted event

The problem would also be present in a world where scheduled command processing is done, although less likely to occur.

So my question is: Does it make sense to "schedule" a series of commands to be published if the saga/aggregate is successfully committed to the event store? Its basically combining an aggregate and synchronous subscriber into one.

ylbillyli:
@rasmus This would be really great. We have been trying to do similar things in our project using aggregate as an event sourced saga with https://github.com/slashdotdash/stateless. Would be perfect if eventflow has native support for saga at the end. Any expected release date so far?

rasmus:
@ylbillyli I have been working on it whenever I got some time, but at the moment I'm a bit stuck. The current implementation has several drawbacks that I don't like, especially regarding how possible errors can be handled.
But if there's an interest, I'll give it another and see if I can find a solution. Also, I'll see if I can post a description of the current problems.

rasmus:
@ylbillyli I found an article that describes what I intend here http://blog.jonathanoliver.com/cqrs-sagas-with-event-sourcing-part-ii-of-ii/ and it should be possible to finish this

ylbillyli:
@rasmus sorry have been busy with project. Nice to see this getting close to live :+1:

PR #118, opened 2015-09-05
Event sourced, try again...

rasmus:
This is a messy cleanup to see if its possible to split the event sourcing part of the aggregate roots into a separate set of classes. This would make it possible to use event sourcing for other parts of EventFlow and not just the aggregate roots.


rasmus:
Will try to do this gradually instead

PR #119, opened 2015-09-06
Added predicate to assembly scanning

rasmus:


Source/EventFlow/Extensions/EventFlowOptionsAggregatesExtensions.cs
```diff
@@ -62,7 +64,9 @@ public static class EventFlowOptionsAggregatesExtensions
             this EventFlowOptions eventFlowOptions,
             IEnumerable<Type> aggregateRootTypes)
         {
-            var invalidTypes = aggregateRootTypes
+            var aggregateRootTypeList = aggregateRootTypes.ToList();
```
JC008: Why do you use a ToList() - its not needed?

JC008: You misses an inline return at Source/EventFlow/Extensions/EventFlowOptionsAggregatesExtensions.cs:59

I will use your style for inline return in future.

rasmus: @JC008 The `ToList()` is to limit the number of times the enumeration is processed. The basic input reads all assembly types, inspects its properties and determines if its of a specific base type. If the `ToList()` isn't there, then all that will be done twice.

rasmus: @JC008 I did a small test by creating 10.000 like shown below after I removed some logging as well. Without the `.ToList()` the results was steady aroud 2.7 seconds, but adding it improved the performance to around 2.2 seconds. I know its minor improvements, but if developers do something crazy in e.g. the new predicate I added, then it would be much worse. Be careful when working with `IEnumerable<T>` and LINQ, you can easily evaluate the expression more than once.

```csharp
[Test]
public void T()
{
    var stopwatch = Stopwatch.StartNew();
    for (var i = 0; i < 10000; i++)
    {
        EventFlowOptions.New.AddAggregateRoots(EventFlowTestHelpers.Assembly).CreateResolver(false);
    }
    Console.WriteLine("{0:0.000}", stopwatch.Elapsed.TotalSeconds);
}
```

JC008: Interesting - thanks.

I found these too. 

http://weblogs.asp.net/lichen/linq-and-ienumerable-may-surprise-you
http://programmers.stackexchange.com/questions/194083/why-should-i-use-listt-over-ienumerablet

PR #120, opened 2015-09-08
Release v0.13

rasmus:


rasmus:
Secure variables aren't decoded for pull requests and thus RabbitMQ and Elasticsearch integration isn't executed... will have to merge manually :cry:

PR #122, opened 2015-09-09
Jobs

rasmus:
Very ugly first implementation of jobs to get a feeling if its something that really should be added to EventFlow.


Source/EventFlow.Hangfire/Integration/HangfireJobScheduler.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+using EventFlow.Core;
+using EventFlow.Jobs;
+using Hangfire;
+
+namespace EventFlow.Hangfire.Integration
+{
+    public class HangfireJobScheduler : IJobScheduler
```
rasmus: @tabro Didn't you have some experience with Hangfire? Would this do for a potential job scheduler?

tabro: @rasmus I have no personal experience - I dont remember who from the team did the implementation though.

PR #125, opened 2015-09-09
Serializable commands... again...

rasmus:
Commands needs to be easily serializable as they basically are value objects


rasmus:
A very bad idea, there are way better ways of doing this

PR #132, opened 2015-09-21
Add configuration modules

rasmus:


rasmus:
@JC008 I added the configuration module support you suggested in #98. What do you think?

JC008:
That will work very well and is similar to [Autofac Modules](http://autofac.readthedocs.org/en/latest/configuration/modules.html) and the comments at the end of the Autofac module documentation re [Common Use Cases](http://autofac.readthedocs.org/en/latest/configuration/modules.html#common-use-cases-for-modules) applies here.

PR #134, opened 2015-09-23
Make "CreateResolver" optional

### LINKED ISSUE #133: Provide a concept similar to Autofac IStartable
Delay EventFlow initialization until the IoC container is ready instead of doing the operations within the `CreateResolver()` method. This will allow developers to omit the `CreateResolver()` completely and initialize the container at a later time better suited for their application.

rasmus:
To do

- [x] Remove the need to explicitly "initialize" EventFlow
- [x] Cleanup code
- [x] Unit tests
- [x] Integration tests

This fixes #133 as the new `IBootstrap` is introduced


JC008:
@rasmus I tried to implement an Autofac.Module to work around the issue of needing to call an initializer but came up against some odd behavior. You can see the attempt in https://github.com/JC008/EventFlow/tree/optional-create-resolver. I made annotations in the code to point out the issues I found and there are lots of test snippets I tried to find out what is going.

You can run a debug session on AutofacAggregateFactoryTests.cs from line 51 to see the flow of the solution and the resulting issue. 

PS: I suspect it might be an async issue.

rasmus:
@JC008 I took a different approach and got it to work. Delete, delete, delete :smile: I removed the registration types, and greatly simplified the Autofac integration. Autofac has a concept regarding default implementations, so the defaults are merely added in the beginning.

I broke the decorator though and need to find a proper solution for that.

rasmus:
@JC008 Also, there's no longer a need for a `Initialize()` method

JC008:
Awesome - nice cleanup + module support - who can ask for more. Thanks.

PR #135, opened 2015-09-23
fix hangfire typos

luisrudge:


rasmus:
Thanks, that is just embarrassing

PR #136, opened 2015-09-25
ICommand<> defined on Command<,> in EventFlow.Commands

JC008:
- caused a break in project dependant on implicit ICommand<,> cast
- now matches Command<,,> with interface ICommand<,,>


rasmus:
Thanks

PR #139, opened 2015-09-28
Drop all test databases script for MsSql Tests

JC008:
- added to /tools/scripts as it seems more appropriate than /scripts

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,8 @@
 ### New in 0.17 (not released yet)
 
-* _Nothing yet_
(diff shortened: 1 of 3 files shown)


rasmus:
@JC008 I cannot accept this into EventFlow as it manipulates databases other than those created by explicitly EventFlow. If someone uses that same prefix for other test databases, this will delete them.

JC008:
@rasmus Sorry - my mistake - the `TEST__` prefixes are used in my projects and its __my__ project test helpers leaving DB's on failed tests for review - not EventFlow MsSql tests. So its not applicable to EventFlow at all.

rasmus:
No worries

PR #141, opened 2015-09-28
Fix critical bug regarding IStartable

### LINKED ISSUE #140: EventFlow startup exception
I adapted a project to use the configuration architecture in 0.16 and have a showstopper. The following minimal program illustrates the exception:
```c#
internal class Program
{
    private static void Main(string[] args)
    {
        var containerBuilder = new ContainerBuilder();
        var eventFlowOptions = EventFlowOptions.New;
        eventFlowOptions.UseAutofacContainerBuilder(containerBuilder);
        containerBuilder.Build();
    }
}
```

and raises

`System.ArgumentException : The type 'EventFlow.Configuration.Registrations.AutofacStartable' is not assignable to service 'Autofac.IStartable'.`

rasmus:
Fixes #140


JC008:
Fixed - thanks for the quick response.

PR #146, opened 2015-10-02
Any entity should be possible to save using event sourcing

rasmus:


rasmus:
This is down the same path as before...

PR #150, opened 2015-10-07
SQLite event store

rasmus:
- [x] How to ensure optimistic concurrency when committing more than one event. As a reference the MSSQL uses a custom table type and inserts all events in one single `INSERT`


rasmus:
@JC008 I started on a SQLite event store. However, I have never worked with SQLite before so it be way off course. It also needs a lot of work still.

JC008:
@rasmus Have a look at https://github.com/praeclarum/sqlite-net, its the Sqlite wrapper I would prefer to use as it's self contained, portable and provides sql statement building. It also removes the need for Dapper.

rasmus:
@JC008 Thanks, it does look more promising

JC008:
@rasmus Have a look at
https://github.com/praeclarum/sqlite-net/blob/master/src/SQLite.cs#L818 and
https://github.com/praeclarum/sqlite-net/blob/master/src/SQLite.cs#L872

The concepts could help with #22

rasmus:
@JC008 I have had a look at sqlite-net and I have made a few observations

**sqlite-net**
* Cons
 - It depends on `sqlite3.dll` which needs to be downloaded from https://sqlite.org
 - It has a poor async/await implementation as it merely wraps the sync methods with  `Task.Factory.StartNew`

**system.data.sqlite**
* Pros
 - Maintained by the SQLite team itself
 - Is a proper ADO.NET provider making it possible to use standard libs like Dapper

It seems like `system.data.sqlite` is a better solution or have I missed something?

JC008:
@rasmus Your summary is correct - note both require the `sqlite3.dll` and you need to build your own 64bit `sqlite3.dll` if you use `sqlite-net` while `system.data.sqlite` provides both the 32 and 64 bit `sqlite3.dll` in their Nuget package. The choice between the two comes down to minimal dependencies in the event that a developer does not want to use ADO.NET and ORM or Sql data wrangler such as Dapper as it is all included in `sqlite-net`.

JC008:
@rasmus BTW in memory databases are possible with Sqlite (you said you have never worked with Sqlite - so might not know about it) which I plan to use for cached readmodels/projections in the future.

https://www.sqlite.org/inmemorydb.html
https://www.connectionstrings.com/sqlite/

rasmus:
@JC008 Thanks for the feedback. I'll have a look at it when I feel better (currently sick and spend a lot of time in bed) :mask:

JC008:
@rasmus I want to complete the Sqlite support but have a question or two:
 
1. `EventFlow.MsSql` provides generic SQL related services (I did a code review and `TODO: Merge with MsSqlConnection` comments in `EventFlow.EventStores.SQLite` point in that direction)
  * should the package be renamed and classes be renamed to be more generic OR
  * a new `EventFlow.Sqlite` package be created (but will almost be a copy of the code in  `EventFlow.MsSql`) OR
  * a new package be created with a more generic `SQL` focus?
2. There has been quite a few changes since Oct 15 and I have tried 3 x `rebase` `develop` on `sqlite` but don't feel confident about the outcomes and have discarded them
  * should I continue with the branch as is and leave merging to later OR
  * will you have time to do a rebase?

rasmus:
@JC008 I din't think that anyone uses the `EventStore.MsSql` package directly so it should be safe to do a rename. Generally I wouldn't do it, but in this case I think it would be the best solution in the long run. Agree?

I have been going through some of the ofter PRs and merging them and I'll get this updated as well.

rasmus:
@JC008 On second thought it might be that it would be better to let the SQLite package have its own usage of Dapper and IConnection for now as there might (I don't know what) be small differences.

rasmus:
@JC008 I did a merge, not sure its all done or even working as I can't work any more on it tonight... but thought that something is better than nothing. It _should_ work though and I added the build configuration for creating the package as well which then can be downloaded from the Appveyor build if you want to test it locally.

JC008:
@rasmus Thanks for the merge and the paket updates - all works fine as long as the packet extension is installed for VS.

I will create a standalone `EventFlow.Sqlite` based on the `EventFlow.MsSql` which can be used to review the potential upside or downside for a consolidated `EventFlow.Sql` in the future.

rasmus:
@JC008 You shouldn't require the Paket extension for VS. NuGet packages should have been installed automatically.

Have you installed the `git lfs` (https://git-lfs.github.com/). The `.paket\paket.bootstrapper.exe` file is downloaded through that.

JC008:
@rasmus I have `git lfs` installed and did not see the bootstrapper pulled down and used the VS addin. I suspect an issue with SourceTree (possible internal or external git installation mixup) and also had issues when you added the icons to the project last year.

rasmus:
@JC008 II use SourceTree as well and the git lfs works great there

rasmus:
@JC008 Is there a deadline for when you would like to have this done? You mentioned that you needed it for a project.

JC008:
@rasmus I will need it by mid Feb at the latest.

rasmus:
@JC008 No worries then, I thought you needed it early next week. I'll see if I can get the SQL package mess cleaned up and merged to develop tonight. And if you aren't working on the SQLite part yet, then I'll merge develop into that branch.

However, I will need some help doing the `SQLiteEventPersistence` implementation correctly and as well as the handling of transient errors as I have absolutely no experience working with SQLite. So, if possible, I would really like if you could have a look at those.

JC008:
@rasmus OK I will wait for the `EventFlow.Sql` implementation and start on the code for the `event` and `read` stores for `SQLite`.

rasmus:
@JC008 If the build is green then only the retry strategy and proper handling of event inserts that are missing and I would really like some help there.

If you want to create read store support as well, then I think that should be done in a separate PR after this is done

JC008:
@rasmus A tool to work with SQLite databases (I use the Essentials edition for Sqlite and Postgresql)

[Navicat Essentials](http://www.navicat.com/products/navicat-essentials)

rasmus:
@JC008 I renamed the SQLite NuGet package to just `EventFlow.SQLite` which then can have support for both event store, read store and the upcoming snapshot store as well. There's no need for all those NuGet packages as the NuGet dependencies are the same. If I didn't do this, then there would be four NuGet packages just for the SQLite part (read store, event store, snapshot store and a common), better make it a bit simpler.

I'll see if I can combine the MSSQL NuGet packages also sometime in the future.

rasmus:
@JC008 Thanks for the link to the tool, but I don't have that amount of money to throw at a tool :cry:

rasmus:
@JC008 Sorry for the spam. But could you have a look at it? I finish the SQLite package as best as I could manage, especially have a look at the `SQLiteErrorRetryStrategy`

rasmus:
@JC008 I haven't heard from you so I'll merge this to develop

jcgenussoft:
@rasmus Sorry - been very busy - will look at it next week

JC008:
@rasmus Sorry - been very busy - will follow up later

rasmus:
@JC008 No problem, its very understandable. It can be hard to find time for open source projects.

PR #153, opened 2015-10-13
DDD example using EventFlow

rasmus:
Create a larger example using EventFlow. The example should

- Give a good understanding of how to use EventFlow
- Give a better understanding of how API changes in EventFlow affect existing applications
- Provide a platform for DDD discussions

The example is based on the shipping example from the book "Domain-Driven Design - Tackling Complexity in the Heart of Software".


JC008:
@ramus (follow on from Gitter) I saw that you where [experimenting](http://martinfowler.com/bliki/TwoHardThings.html) (namespaces can be really tough) e.g. the readmodel placement at the root and an Application level service definition + module level service definition. If you assume/accept/decide that a bounded domain lives in one assembly it affects the namespace usage in the assembly - for me there should be a `Models` namespace at root so you end up with `EventFlow.Examples.Shipping.Models.Cargo`. 

Currently I plan on partitioning out the domain interfaces in my implementations- and then partitioning the implementations further per storage/technology provider for example from your example domain I would implement `Shipping.Cargo.dll`, `Shipping.Cargo.Provider.dll`, `Shipping.Cargo.Provider.MsSql.dll`, `Shipping.Cargo.Provider.RabbitMq.dll` and use the EventFlow module system to register the appropriate parts 

`Shipping.Cargo.dll` - only domain specific interfaces e.g. IRoute, ICargo, ILocation (maybe commands too - iow anything with external exposure e.g. ICargoBook)

`Shipping.Cargo.Provider.dll` - anything that does not use an event store and is 3rd party technology agnostic e.g. Events, Commands, Specifications, ValueObjects, Entities

`Shipping.Cargo.Provider.MsSql.dll` - MsSql implementation of event storage and read models

`Shipping.Cargo.Provider.RabbitMq.dll` - RabbitMq event broadcasting

The approach allows for changing readmodels and storage without affecting the core domain implementation.

EventFlow is flexible enough to handle it I think, but may its to much? 

Your thoughts?

rasmus:
@JC008 I would really like to have the domain itself be completely unaware of read models and will pull them into another project at some time. My basic idea is only to have queries that return either immutable value objects or immutable entities in the domain project. Then how the query handlers are implemented would be put in a different project. I would then create e.g. two separate implementation projects for the query handlers e.g. like this

- EventFlow.Examples.Shipping.Domain
- EventFlow.Examples.Shipping.Domain.Queries.InMemory
- EventFlow.Examples.Shipping.Domain.Queries.Elasticsearch

The in-memory queries would then be used for all integration tests of the domain enabling a much faster execution of tests. Explicit tests of each query handler in the Elasticsearch project will then be done separately.

Basically for me time is very important and the time it takes for me to execute tests is wasted. I would like to be able to use e.g. NCrunch or the new integrated testing feature in ReSharper 10 on integration tests and thereby get instant feedback while I'm working on a bug fix or new code. For that the tests needs to be fast.

Also, read models doesn't really seem like a part of a domain as they are in their core just an optimization.

This is just ideas at the moment and I'm all ears for any suggestions as its also one of the main objectives of creating the example.

Note that I push everything I do to GitHub, even the smallest change. This will let people looking at the EventFlow to get a feel of what is happening, but I will also leave a lot of the pull requests in a somewhat bad state. The example, is barely working and a lot of work is still needed.

What you mention about project structure makes good sense and also hooks into these service concepts

- Application services, i.e., how an application interacts with the domain
- Domain services
- Infrastructure services, e.g. sending mail

I like your suggestion about splitting the bounded contexts into separate projects.

PR #160, opened 2015-10-23
Added a resolver to the read model context

rasmus:


JC008:
@rasmus Excellent timing - I just ran into the problem of not being able to DI anything on a readmodel. I use the following type in EventFlow readmodel events for lookups with a temp global resolver.

```c#
public interface ILookup<in TKey, TValue>
{
    TValue Lookup(TKey key, CancellationToken cancellationToken = default(CancellationToken));
    Task<TValue> LookupAsync(TKey key, CancellationToken cancellationToken);
}
```
Also - I seem to be populating `Entity<>` objects from the readmodels in my implementation - very interesting - you still have DDD entities but they are managed by EventFlow and source from the ES,

rasmus:
@JC008 I'll make sure to release it early next week then.

I also did something similar for the shipping example with creating immutable `Entity<>` obejcts based on read models.

https://github.com/rasmus/EventFlow/blob/example-shipping/Source/EventFlow.Examples.Shipping.Queries.InMemory/Cargos/CargoReadModel.cs#L57

PR #163, opened 2015-10-24
Remove references to aggregates from event persistence

rasmus:
In preparation of separating the event sourcing parts of EventFlow, the event persistence layers needs to have all references to aggregates removed. The underlying data models of e.g. the MSSQL event store will still have columns that have these names to still be backwards compatible.


rasmus:
Doh

PR #166, opened 2015-11-03
Crude PoC on how to use read model locators with MSSQL

rasmus:
This is merely to show how read model locators could be used with MSSQL. Don't merge it!


rasmus:
Prettier example found in #170

PR #169, opened 2015-11-09
Better read model documentation

rasmus:
Almost everything is better than nothing...


Documentation/ReadStores.md
```diff
@@ -0,0 +1,109 @@
+# Read model stores
+
+In order to create query handlers that perform and enable them search across
+multiple fields, read models or projects are used.
```
JC008: projects > projections

Documentation/ReadStores.md
```diff
@@ -0,0 +1,109 @@
...
+
+Read models are a flatten views of a subset or all aggregate domain events
```
JC008: flatten > flattened

Documentation/ReadStores.md
```diff
@@ -0,0 +1,109 @@
...
+created specifically for efficient queries.
```
JC008: + by denormalizing the stored projection to avoid the cost of table and index scans when joining tables in sql.

Documentation/ReadStores.md
```diff
@@ -0,0 +1,109 @@
...
+
+Here's a simple example of how a read models for doing searches for usernames
```
JC008: models > model

Documentation/ReadStores.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+## Read model locators
+
+Typically the ID of read models are the aggregate identity, but sometimes this
+isn't the case. Here are some examples.
+
+- Items from a collection on the aggregate root
+- Deterministic ID created from event data
+- Entity within the aggregate root
```
JC008: - root

Comment: referring to `aggregate root` which is conceptually an entity could be confusing or is the intention to state that an entity can have an entity as dependency?

rasmus:
@JC008 Thanks

PR #170, opened 2015-11-09
More freedom when using MSSQL read models

### LINKED ISSUE #167: Remove enforced columns for MSSQL read store
The MSSQL read store should not enforce columns on read model tables, developers should be free to create any schema for their read models, similar to the Elasticsearch implementation.

rasmus:
Fixes #167


Source/EventFlow.MsSql.Tests/IntegrationTests/ReadStores/ReadModels/MsSqlThingyMessageReadModel.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace EventFlow.MsSql.Tests.IntegrationTests.ReadStores.ReadModels
+{
+    [Table("ReadModel-ThingyMessage")]
+    public class MsSqlThingyMessageReadModel : IReadModel,
+        IAmReadModelFor<ThingyAggregate, ThingyId, ThingyMessageAddedEvent>
+    {
+        public string ThingyId { get; set; }
+
+        [MsSqlReadModelIdentityColumn]
```
JC008: Nice clean backwardly compatible zero impact solution to locating the key field :+1:

rasmus: Thanks, still needs work though.

JC008:
@ramus Is it possible to make the `ReadModelSqlGenerator` more usable for class inheritance by marking the sql dictionary members as `protected` and not `private`? Reason: I need to add custom recursive CTE queries as select sql for specific read models and would like to extend `ReadModelSqlGenerator` and use it with a specialized `IMssqlReadModelStore` implementation.

rasmus:
@JC008 If the changes aren't needed within the EventFlow read model update process, then I would recommend you to use aggregation instead of inheritance as It will give be more control. If the changes are required during the read model updates I would still recommend you to use aggregation and then apply it to EventFlow using the decorator interface. That way you will get some nice compile errors if new methods are added or changes are made to methods that you don't override.

Issue #165, opened 2015-10-31
SqlReadModel handling for aggregate collection data


JC008:
@rasmus How do you handle creating read models for the following scenario using the `MsSqlReadStore` (or any `SQL` based read store):

- the aggregate is a job order (e.g. repair something)
- the job order aggregate has line items in a collection
- line items may be *owned* by other line items 

The job aggregate read model is not an issue as it maps to the `IMsSqlReadModel` and handles events via `IAmReadModelFor<,,>` with SQL generated targeting the `AggregateId` field in the table by `EventFlow.ReadStores.MsSql.ReadModelSqlGenerator`. 

A `ReadModelSqlGenerator` cannot be registered on a specific read model, only a `IReadModelLocator` can and in this case the read model requires specific SQL to select the line item.

How are the line items handled?

rasmus:
@JC008 Just to make sure that I understand the scenario. The job order is the entity implemented as the aggregate root, and the line items are basically entities in a collection on the job order.

If that's the case, then you could implement the read model like you would normally would, then implement an `IReadModelLocator` that merely returns the ID of the specific line item. So if you have a set of entity instances like this:

* job order with ID `1`
* line item with ID `a` belongs to job order `1`
* line item with ID `b` belongs to job order `1`

then the `IReadModelLocator` would return `a` and `b` for the domain events that are specific for those line items.

If that fails, then implement the read model logic manually using the `ISubscribeSynchronousTo`<,,>` interface.

If that doesn't solve your problem, or you have a better idea, let me know.

JC008:
@rasmus Here is another scenario matching the shipping example.

Cargo (or PAX) is never shipped without a manifest, the details of what the cargo contains as well as its international classifications as well as width, height, weight, value etc + [export duty codes](https://en.wikipedia.org/wiki/Harmonized_System)  e.g.

* 001 - 10 boxes of bananas - 2kg / 200x200x300, store below 10C in dark - IATA code XYZ1
* 002 - 3 crates of high explosives - 20kg / 2000x1000x1000 - IATA code ABS1

To update the read model for an item in the event that the manifest changes, the `IReadModelLocator` will return the correct id for the manifest entry (e.g. `001` to update the babanas entry) but the SQL will be `SELECT * FROM CargoManifest WHERE AggregateId = 001` and will not find the item. 

This is a case where an `IIdentity` (e.g. CargoItemId) is used but links to a collection in an aggregate and the readstore only targets aggregates.

[...]

PR #179, opened 2015-12-05
Fix: See if there's a problem regarding multiple emits and read models

rasmus:
Related to the discussion in #165

### DIFF
--- a/Source/EventFlow.TestHelpers/Aggregates/Commands/ThingyImportCommand.cs
+++ b/Source/EventFlow.TestHelpers/Aggregates/Commands/ThingyImportCommand.cs
@@ -1,0 +1,71 @@
+﻿// The MIT License (MIT)
+// 
+// Copyright (c) 2015 Rasmus Mikkelsen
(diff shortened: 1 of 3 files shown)


JC008:
@rasmus My issue in #165 can be illustrated using the `ThingyMessageLocator` as example:

```c#
public class ThingyMessageLocator : IReadModelLocator
{
    public IEnumerable<string> GetReadModelIds(IDomainEvent domainEvent)
    {
        var messageAddedEvent = domainEvent as IDomainEvent<ThingyAggregate, ThingyId, ThingyMessageAddedEvent>;
        if (messageAddedEvent == null)
        {
            yield break;
        }

        // -> omitted by me - yield return messageAddedEvent.AggregateEvent.ThingyMessage.Id.Value;
    }
}
```

I missed that the locator **MUST** return a result.

rasmus:
@JC008 There was a bug, I fixed in #179

rasmus:
@JC008 Sorry, its fixed in #180

PR #187, opened 2016-01-22
Fix WrongExpectedVersionException issue

oliveiracdz:
Hi! 

The EventStore stream version is zero based index: -1.
The expected version is the previous version: -1 again.

It's my first PR, please sorry if i did something wrong...

Thank you! =)


rasmus:
@oliveiracdz No worries, its good to get it fixed. I haven't had enough focus on EventStore yet, so thanks for creating the PR.

Could you add a unit test (if possible) and update the release notes as well. I'll release EventFlow when this gets merged to get the bug fix out.

rasmus:
Fixed in #190, which is based on this PR.

PR #189, opened 2016-01-24
Wrong expected version in EventStoreEvent

rasmus:
Based on PR #187 from @oliveiracdz


rasmus:
@oliveiracdz I created a integration test and executed it locally. I'll merge this, create a release and close #187 

I'm struggling a bit with the automated tests in #190, but I'll see if I can get them stable so they become a part of the build process.

PR #190, opened 2016-01-24
Enable EventStore integration tests

rasmus:
Automatically download and run a EventStore instance for the integration tests


rasmus:
This provides better testing for #187

rasmus:
Darn

rasmus:
Tests are unstable as they sometimes fail with this exception

`EventStore.ClientAPI.Exceptions.CommandNotExpectedException : Expected : ReadStreamEventsForwardCompleted. Actual : ReadStreamEventsForward`

I cannot reproduce it locally and it isn't the same test that fails.

rasmus:
It seems like the stability problems in the EventStore integration tests were due to TCP self connects. The simple solution was to use the same EventStore server for all integration tests

PR #196, opened 2016-01-31
Create EventFlow.Sql for generic SQL services

rasmus:
@JC008 I created this branch to implement the rename of `EventFlow,MsSql` and made changes that enables it to be used by e.g. SQLite in #150.


JC008:
@rasmus I have started on `EventFlow.SQLite' for #150 and need to check if the error codes in `SqlErrorRetryStrategy` are universal across ADO providers or if SQLite specific code must be used.

rasmus:
@JC008 I made it possible for SQLite to have its own retry strategy and configuration in the PR as I didn't think the error codes were common

rasmus:
@JC008 I you think this solution would work, then I'll get it done and get it to `develop`. Then, unless you are doing changes locally, I'll merge develop into the SQLite branch.

Source/EventFlow.Sql/Integrations/TableParameter.cs
```diff
@@ -31,7 +32,7 @@
 using Dapper;
 using Microsoft.SqlServer.Server;
```
JC008: Dependency on `Microsoft.SqlServer.Server` should not be allowed.

rasmus: @JC008 Very, very good point... Then I'll have to reintroduce the EventFlow.MsSql package once again and move the table parameter to that.

JC008:
@rasmus Maybe its worth while adding another level of abstraction now to enable implementations (for `ReadStores` and `EventStores`) with NoSQL and Cloud storage systems later (e.g. Azure table store)?

rasmus:
@JC008 I don't follow that abstraction would be? There's already implementations for Elasticsearch as a read store and EventStore as another event store and those don't rely on anything but the core EventFlow package and the specific client libraries.

JC008:
@rasmus OK - ignore that thought then - I have not worked with those implementations yet.

PR #198, opened 2016-02-10
Use TinyIoC instead of IL merge Autofac in core package

rasmus:
Well, the title says it all...

IL merge has proven troublesome in the past, so I'll try to get rid of it completely and the first step is to remove the dependency on Autofac from the core package.

This PR will see if its is indeed possible to get TinyIoC to work with EventFlow.

Note: The EventFlow.Autofac package will not be removed!


rasmus:
Not all required features are supported :cry:

Issue #205, opened 2016-02-23
Custom Hangfire Job Title?


ylbillyli:
I have been using the IJob and IJobScheduler to queue jobs on Hangfire, but couldn't get the Job Name displayed right on the hangfire dashboard even with the [DisplayName] attribute or overriding the ToString() method on the Job. Is this supported by EventFlow?

![image](https://cloud.githubusercontent.com/assets/13381578/13244758/b9e99a58-da43-11e5-8b7c-3e530282f757.png)

rasmus:
@ylbillyli You are right, that would be supported by EventFlow and I'll look into it.

rasmus:
@ylbillyli The #206 is now live in release v0.27.1765

ylbillyli:
Thx @rasmus I believe #206 will close this off

PR #206, opened 2016-02-23
Allow developers to control job display names in Hangfire

rasmus:
Developers can implement a custom `IJobDisplayNameBuilder` and control how a job display name is created.

@ylbillyli Would something like this work for #205?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,10 @@
-### New in 0.26 (not released yet)
+### New in 0.27 (not released yet)
+
+ * New: Configure Hangfire job display names by implementing
+   `IJobDisplayNameBuilder`. The default implementation uses job name and
+   version.
+
+### New in 0.26.1714 (released 2016-02-20)
 
  * Breaking: Renamed `MssqlMigrationException` to `SqlMigrationException`
  * Breaking: Renamed `SqlErrorRetryStrategy` to `MsSqlErrorRetryStrategy`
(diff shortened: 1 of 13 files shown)


ylbillyli:
So we just have to implement the IJobDisplayNameBuilder and then can build the job name ourselves? Nice solution. In my own project, I have tried to create new IJobRunner interface myself with the DisplayName attribute and then replaced the IJobScheduler implementation to use my IJobRunner to get around. But would prefer to upgrade to this when it is available.

ylbillyli:
Ahh, one more problem, best if we could make `JobDisplayNameBuilder.GetDisplayNameAsync` generic to customise individual job name which then would be perfect.

something like 
`GetDisplayNameAsync<TJob>(TJob job, JobDefinition jobDefinition, CancellationToken cancellationToken )`

Which then I can build the job name using properties in job like 
`return $"{jobDefinition.Name} for Client {job.ClientId}"'

rasmus:
@ylbillyli Making it `GetDisplayNameAsync` generic would not make it easier as the service would still be resolved once as the generic isn't on the interface. Making the interface would force you to register a name build for every job and I would argue that this would be too cumbersome compared to simple implementing one.

The idea is, that `GetDisplayNameAsync` allows you to build whatever you would like to use for job names. The default will use the job definition name and version and thus providing a sensible default.

If you choose to implement your own, then you could inspect the job properties or select to simple do the convention that the job title is the `ToString()` of a job.

rasmus:
image

ylbillyli:
Cool. `GetDisplayNameAsync` takes the IJob. I wasn't sure that the overridden ToString() on the actual job can still be called even it is converted to the interface IJob inside `GetDisplayNameAsync`. Just now linqpad proved it does work! I agree yours would be one of the cleanest solution to my problem.

PR #230, opened 2016-06-20
Allow unhandled subscriber exceptions to be thrown

rasmus:
Sometimes unhandled subscribers exceptions should be allowed to propagate to the originating publisher of a command.


Source/EventFlow/Subscribers/DispatchToEventSubscribers.cs
```diff
@@ -83,8 +105,8 @@ private class SubscriberInfomation
             {
                 _log.Error(
                     exception,
-                    "Failed to dispatch to event handler {0}",
-                    handler.GetType().PrettyPrint());
+                    $"Failed to dispatch to event handler '{handler.GetType().PrettyPrint()}' for event '{domainEvent.EventType.PrettyPrint()}' due to unexpected exception!");
+                throw;
```
ylbillyli: How about save one try-catch overhead by doing it here:
```
if(_eventFlowConfiguration.ThrowSubscriberExceptions) throw else _log.debug(e, "swallowed");
```
Just a small suggestion, not sure if I am up-to-date with the code base, if you are doing an extra `try-catch` in `DomainEventPublisher` on purpose to make it neater, then I think that is also fair enough.

rasmus: @ylbillyli You are right. I was thinking that I wanted to preserve the previous logging behavior, but a cleaner codebase is more favorable

rasmus:
@ylbillyli I did some changes based on your comments. It cleaned it up a bit

ylbillyli:
@rasmus looks perfect to me

PR #231, opened 2016-06-21
Upgrade to Elasticsearch 2.x

rasmus:


rasmus:
Delete by query was deleted in 2.x https://www.elastic.co/blog/core-delete-by-query-is-a-plugin

PR #234, opened 2016-06-29
.NET Core - part 1

rasmus:


rasmus:
http://www.michael-whelan.net/porting-dotnet-framework-library-to-dotnet-core/

riezebosch:
Was eager to try out your package but current release has a dependency on net451. Any progress on the .NET Core support?

rasmus:
@riezebosch No progress at the moment 😢 and I'm a bit unsure on how to proceed as the project.json is deprecated (https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/)

If you can point me in the right direction I'll see what I can do.

riezebosch:
The changes on the SDK make it [a bit of a pain](https://weblog.west-wind.com/posts/2016/Nov/23/NET-Standard-20-Making-Sense-of-NET-Again#.CSPROJ-It'sbaaack!!!) at the moment indeed. But the compiled assemblies should be compatible AFAIK so porting the source code would still make sense. The only problem on that side is that with netstandard2.0 around the corner the API surface will increase which will [make porting easier](https://weblog.west-wind.com/posts/2016/Nov/23/NET-Standard-20-Making-Sense-of-NET-Again#.NETStandardVersions) in the near future.

rasmus:
@riezebosch So from what I can read the recommendation is to wait for VS2017 and .NET Standard 2.0

riezebosch:
I created a [pull request](https://github.com/eventflow/EventFlow/pull/278) to notify you the progress I've made while porting EventFlow. You probably shouldn't accept that request right away but you could take a look at the changes.

In the mean time I also ported the EventFlow.Autofac project, but I cannot port the test projects due to the dependency on EventFlow.TestHelpers which has a dependency on AutoFixture. On the AutoFixture project the exact same discussion takes place on whether to port to dotnetcore already: https://github.com/AutoFixture/AutoFixture/issues/404

rasmus:
Better way in #291

PR #235, opened 2016-06-30
Read model optimistic concurrency

rasmus:


houseme-kyle:
Any idea on the ETA for this pull request? Happy to assist if there are specific TODOs that can be done.

rasmus:
@houseme-kyle no ETA yet, but I am slowly starting up again on EventFlow and I'll will get more time in the next weeks. I'll try to get this done first.

Wouldn't you be more interested in this PR instead https://github.com/eventflow/EventFlow/pull/245 ?

houseme-kyle:
Could very well be that one causing us difficulties. Commented on this one as I saw it first, but what we seeing are updates being lost when two events are applied simultaneously.

rasmus:
@houseme-kyle Both require that versioning is supported by your read model. If the version is on the read model itself you could simply compare that with the aggregate sequence number of the domain events to be applied and log if there's a mismatch. A built-in feature would be best, but this would enable you to get an idea of the impact

rasmus:
@houseme-kyle just disregard my previous comment, it doesn't really make sense and I can see that this is a prerequisite for PR #245 and thus I'll get this done asap.

houseme-kyle:
Thanks @rasmus feel like this is one of the last features to make the read model production ready.

rasmus:
Hi @houseme-kyle, I picked this up again. Do you have any comments regarding the code at this point? If not, then I'll see if I can merge it in this week

PR #245, opened 2016-08-23
WIP: Apply missing events to read models

rasmus:
If a race condition occurs for read model updates for a read model update gets lost, we need to bring the read model up to date.


Source/EventFlow.MsSql/ReadStores/MssqlReadModelStore.cs
```diff
@@ -142,6 +142,7 @@ static MssqlReadModelStore()
                     readModelEnvelope,
                     cancellationToken)
                 .ConfigureAwait(false);
+            if (readModelEnvelope == null) return;
```
rasmus: @frankebersoll I would like to finish this PR soon, but I'm very unsure how well I managed to do the implementation, especially this part (return null signals no updates to read models). If you have the time I would very much like a second pair of eyes on this.

rasmus: Same here. I think I'll go with your suggestion regarding the `ReadModelUpdateResult` as this part will closely integrate with read model stores and being explicit here makes it easier to create integrations instead of having some secret meaning to specific values.

Thanks for the input,

rasmus: Trying a `IsModified` on the envelope

frankebersoll: I'm not sure... The envelope is also used "externally" by IReadModelStore users, mainly query handlers. I won't expect an "IsModified" on the envelope there either.

rasmus:
I'm currently focusing on closing old PRs that many have requested. If anyone want to help out, then getting the tests green here would be great help

Source/EventFlow.MsSql/ReadStores/MssqlReadModelStore.cs
rasmus: @frankebersoll Didn't see your response, sorry. Will have a look

Issue #51, opened 2015-04-28
Consider change to LibLog


rasmus:
Instead of creating yet another log framework within EventFlow, consider changing to LibLog and then embed the DLL with ILMerge.

https://github.com/damianh/LibLog

rasmus:
Maybe just add a extra logger called `LibLog` that developers can register

JC008:
A good write up on the subject - [Which logging abstraction should I use?](https://nblumhardt.com/2016/04/which-logging-abstraction-should-i-use/)

rasmus:
@JC008 Thanks

PR #258, opened 2016-10-19
Init LibLog

AntoineGa:
#51

Added LibLog
Modified Method to log


AntoineGa:
The build fails as i don't know which solution to pick for defined logger inside EventFlow:
-should you want to use a configuration from popular logger (Serilog ? NLog ?) for ConsoleLogger / NullLogger 
-Shoud i implement all the LibLog method for  ConsoleLogger / NullLogger and 2 simple method to config 
?

rasmus:
@AntoineGa Thanks for the PR, I have some input

- `package.config` was added, please remove it and use Paket instead of NuGet (https://fsprojects.github.io/Paket/)
- Several verbose logging has been changed to information, these need to be changed back. The level of severity order is verbose, debug, information, warning and then error
- The references to the `EventFlow.Logs` namespace has changed to `EventFlow.Logging`, but no files have been addeed or moved
- I would prefer if the change to LibLib was done in two steps, first a release which enables a `UseLibLog()` on `IEventFlowOptions` that implements a new logger in EventFlow that uses LibLog and thereby changing no interfaces at all. LibLog should be included in the assembly and not refereced by the NuGet package (its one of two usages of LibLog). Then at some later release, `ILog` _could_ be removed and LibLog used directly

rasmus:
@AntoineGa Also, please revert `.paket/paket.bootstrapper.exe`. I would like to download that directly myself if its updated

AntoineGa:
the LibLog does not define Verbose level but a Trace instead.
`    enum LogLevel
    {
        Trace,
        Debug,
        Info,
        Warn,
        Error,
        Fatal
    }`

That's why i have modified Verbose but i should have mapped them to Trace

wgtmpeters:
@rasmus  will this ever be merged?
would help our project greatly

rasmus:
Implemented in #428

PR #261, opened 2016-11-02
Update Autofac* packages to 4.x

rasmus:


rasmus:
no need to update, 3.x seems compatible with 4.x, will loosen up dependency instead

PR #265, opened 2016-11-07
Update GettingStarted.rst

accoleon:
Under the "Create command handler" section, the implemented base class should have been CommandHandler instead of ICommand


rasmus:
@accoleon Thanks

PR #278, opened 2016-11-28
.NET Core port with .NET451 compat

riezebosch:
I managed to port the EventFlow project to .NET Core. All other projects are still .NET 4.5.1 to prove that everything still works. 

* Replaced the EventFlow.csproj with a EventFlow.xproj (for now). 
* Replaced project referecences from all other projects with an assembly reference to the output directory of EventFlow because project references from a *.csproj to a *.xproj doesn't work. This probably will probably work again with VS2017 and .NET Core SDK Preview 3.
* Replaced all reflection methods with the .NET Core API. Mainly there is an additinal GetTypeInfo on almost every Type method because that information is no longer available directly on the Type class. There may be a lot of refactoring possible on this part!
* Replaced the SHA methods with the API available in .NET Core. There weren't unit tests targeting this method directly but I found out it is covered with the identity tests.

This is mainly a proof that it will work. To be useful all other projects should also be ported. My guess is this port is future proof because there aren't that many API methods that you use and aren't supported yet. The main difference will be the migration back to *.csproj files but then again these can be migrated using the tooling.

When you run a `dotnet pack` in the EventFlow folder you get a nupkg containing a library that targets .net451 and one that targets dotnetstandard1.6.


rasmus:
@riezebosch Very nice, I'll have a look at it as soon as I get settled in the new house. Everything is still a bit of a mess

rasmus:
Closed in favor of #303 which uses the new VS2017 project format

PR #295, opened 2017-01-08
Internal IoC container

### LINKED ISSUE #142: Internal Autofac stability
Either write a test that ensures that the IL merge of Autofac doesn't break when the NuGet package is used. That, or remove the need to IL merge Autofac.

rasmus:
Want to remove the IL merge of Autofac for the core package and don't want to have it as a dependency. 

- Makes it easier to provide .NET Core packages
- Simplifies the build and package process
- Remove old tool, ILMerge hasn't been updated since 2014

Also fixes #142 by making it obsolete.


rasmus:
Useful 

- http://www.palmmedia.de/blog/2011/8/30/ioc-container-benchmark-performance-comparison
- https://gist.github.com/vcsjones/716137

PR #298, opened 2017-01-16
File system wrapper

rasmus:
Introduce a thin file system wrapper. Will be used later for e.g. opt-in event stream archives.


rasmus:
Will be implemented in #317

PR #304, opened 2017-04-05
Allow a more "eventually consistent" programming model

rasmus:
Allow developers to configure whether or not they wish to wait for all domains events to be published. This allows developers to have a more eventually consistent model as the command publish ensures that the events are committed, but the read models and subscribers might not be updated.


rasmus:
As #335 mentioned, using the task runner yields unexpected results... need to rethink this

PR #309, opened 2017-04-10
PoC - Remove Paket

rasmus:
With all the changes made to Visual Studio and NuGet, lets check to see if its time to move away from Paket and simply use native tools as more will be familiar with these.


rasmus:
Made obsolete by #324

PR #317, opened 2017-04-15
WIP: Event archives

rasmus:
When a event store becomes too large, we would like the option to move some of the events to a different location.


rasmus:
It seems like it would be optimal to wait for async streams and `IAsyncEnumerable` in C# 8 instead of inventing a wrapper

rasmus:
Too old

PR #319, opened 2017-04-27
Create test scenario for MVC5

rasmus:


rasmus:
Use later version instead

PR #328, opened 2017-05-03
Pull in the Helpz required classes

rasmus:
After moving support classes to Helpz it has become a bit of a hazel when working with MSSQL tests as any changes always required additional changes to an external package.


rasmus:
Makes iterating on #329 easier

PR #330, opened 2017-05-03
.NET Standard 1.6 support - take 2

rasmus:


pvivera:
@rasmus Excelent job! I'm working to add Postgres support but I want to do it with netstandard in place. I'll be using EventFlow.Sql, like you did it for MsSQL and SQLite, but it has a dependency on dbup and it doesn't support netstandard. I found that there is there a beta version suppoted (dbup4-core, dbup4-sqlite and dbup4-sqlserver) but I don't if you're open to use it.
Please tell me your thoughts about it.

rasmus:
The problem with using pre-release packages is that any EventFlow packages will also have to be marked as pre-release. I know that I haven't released a 1.x yet, but marking a package as pre-release is something else.

Alternatively the dbup parts could be removed from the .NET standard assembly by removing the `SqlDatabaseMigrator` with some `#if NET451` until dbup is ready. This will leave developers with the problem of implementing their own, but it would enable a .NET standard package.

rasmus:
@pvivera have a look at #334. would something like that make sense? you could then in your project reference the unofficial package and copy-paste the `SqlDatabaseMigrator` implementation

pvivera:
@rasmus It looks really clean to me.

PR #332, opened 2017-05-09
Allow lower versions of Autofac

rasmus:


rasmus:
NET Standard isn't supported in the older versions

PR #336, opened 2017-05-22
Use job scheduling for asynchronous subscribers

### LINKED ISSUE #335: Tasks are terminated in web API
Which is currently only used by the opt-in `ISubscribeAsynchronousTo<>`

See http://blog.stephencleary.com/2012/12/returning-early-from-aspnet-requests.html

rasmus:
Instead of implementing yet another job/task scheduling, simply use the functionality already in place. Users can then use e.g. `EventFlow.Hangfire` to do proper asynchronous subscribers without having to do anything.

Fixes #335


rasmus:
@tabro this should fix your problem

RELEASE_NOTES.md
```diff
@@ -1,6 +1,23 @@
 ### New in 0.45 (not released yet)
```
rasmus: @tabro Could you have a look at this as you would be directly affected. Sort version, asynchronous subscribers are disabled by default and if enabled are executed using the scheduler, i.e., Hangfire in your case.

tabro:
@rasmus that looks great - I am ready to test it out as soon as you have a release ready 👍

PR #338, opened 2017-05-24
Validate that custom aggregate IDs are possible

rasmus:


rvcssingh:
I wanted to pass two I'd in my let say employee aggregate constructor that eventually pass to aggregate root constructor.first I'd would be aggregate I'd and other one of tenant id.when I put  2 Param in constructor then it says it need to register in process of aggregate creation .do I need to twick in aggregate factory or so.

rasmus:
EventFlow expects an aggregate only to have one identity. You could create your own composite identity, i.e., the identity type having values like `tenant:id`, e.g. `cool product 1:user-1` and `cool product 2:user 1`. In the example the `cool product x` is the tenant and the `user 1` is the ID within that tenant. The identity type could have multiple helper methods that lets you access the tenant and ID easily. Somewhat similar to `Identity<>` in EventFlow which consists of the type name and a GUID

rvcssingh:
How to delete a aggregate. for eg. let say I have added one employee now I need to delete it by raising employeeDeleted event

PR #364, opened 2017-08-30
Build XML documentation

rasmus:
Better late than never.


rasmus:
will add documentation over time

PR #376, opened 2017-09-04
WIP: Graph domain

rasmus:
Investigate if being able to auto-generate graphs of the domain is any use.

graph


rasmus:
@tabro Didn't we discuss something like this? Maybe put in any event subscribers as well. If you have any comments, let me know.

tabro:
@rasmus Im especially interrested when two aggregates is listening in on eachother with subscribers. So YES!

rasmus:
Too old

PR #377, opened 2017-09-06
WIP: Command results

### LINKED ISSUE #333: Command results
Investigate if command results would be ideal for EventFlow

Sources
- https://jimmybogard.com/domain-command-patterns-handlers/
- https://jimmybogard.com/domain-command-patterns-validation/
- https://gist.github.com/vladikk/86da55d0eb09d7a291b9f9a5b406f2c9

rasmus:
Fixes #333


edwardwilson:
@rasmus Looks a good start. I notice that they command handlers are not typed. It seems that this could be tricky due to the many `i.GetGenericTypeDefinition() == typeof(ICommandHandler<,,,>)` checks through out. I think this could be difficult to handle unless all command handlers have to return TResult.

What are your thoughts about making all commands `ICommand<TResult>`?

rasmus:
@edwardwilson I been trying really hard not to make any breaking changes. I know that there isn't a "1.x" release yet, but messing with commands and command handler could be painful for some. I would really like to see this in EventFlow and making the breaking change might be the only option moving forward.

rasmus:
@edwardwilson Did another iteration and allowed a few breaking changes. Should still be easy to get a large solution to compile again due to the friendly overloads. Needs more work though, still seem a bit "hackish".

edwardwilson:
@rasmus I think this looks good. I took a look yesterday myself and found I was doing down a deep hole trying to keep the sourceId in the generics. I think it is cleaner without it. 

Where do you feel there is hacks in this solution? Are there areas which need improving?

rasmus:
@edwardwilson there were some cleanup in the command bus. I think its getting there, merely some additional tests, more cleanup and a good update to the release notes on how to migrate.

The only thing that buggers me somewhat, is the rename of `ExecuteAsync` to `ExecuteCommandAsync` on the command handler interface. Its done to allow easier migration of implementations that inherit from the `CommandHandler<,>` base class. I would like not to do it.

rasmus:
@edwardwilson introduced "execution results"

edwardwilson:
@rasmus think this is looking really good. Exactly what I had in mind for a TResult on commands. It's good that the result is an options thing too.

edwardwilson:
@rasmus Happy to help with documentation but with it being in a new repo now the samples will need to be updates after?

rasmus:
I was thinking about setting up AppVeyor package feed and using that for the documentation repo. It will allow an easier update flow for updating the documentation of new features. I will have some time this weekend and I hope to finish it up. I want to create some scenarios to see how the new interface is to work with. It would be pretty bad to create a breaking chance and then need to do another to fix it.

I was thinking about updating the small site on geteventflow.net, write a blog post about it there and then link to that from the release notes. EventFlow is starting to be used more and more, although not much yet. But I would like to handle breaking changes "professionally".

PR #384, opened 2017-09-26
WIP: Consume domain events from RabbitMQ

### LINKED ISSUE #383: Consume domain events from RabbitMQ
As EventFlow is able to publish domain events to RabbitMQ, it should also be able to consume them if you want to link two bounded contexts using events.

rasmus:
Fixes #383


rasmus:
Too old

PR #385, opened 2017-09-26
PoC: Redis

rasmus:
Might not make much sense


rasmus:
Doesn't make much sense

PR #391, opened 2017-10-12
Issue #357: File store should be thread safe. Only allow a single thread write access to the file event store log.

### LINKED ISSUE #357: File store should be thread safe
To allow easier testing of EventFlow in a setup, the file event store should be thread safe.

See #356

t-johnson:
This implements a semaphore that will serialize all writing to all log.store files, in case there are more than one in use, even though it would be acceptable to be writing to two different log.store files in different folders.

Fixes #357

### DIFF
--- a/Source/EventFlow/EventStores/Files/FilesEventPersistence.cs
+++ b/Source/EventFlow/EventStores/Files/FilesEventPersistence.cs
@@ -40,7 +40,7 @@ namespace EventFlow.EventStores.Files
         private readonly IFilesEventLocator _filesEventLocator;
         private readonly AsyncLock _asyncLock = new AsyncLock();
         private readonly string _logFilePath;
-        private static SemaphoreSlim _logFileSemaphore = new SemaphoreSlim(1,1);
+        private static AsyncLock _logFileSemaphore = new AsyncLock();
         private long _globalSequenceNumber;
         private readonly Dictionary<long, string> _eventLog;
 
@@ -169,8 +169,7 @@ namespace EventFlow.EventStores.Files
                 }
 
                 //only allow a single write access to the _logFilePath file at one time
-                await _logFileSemaphore.WaitAsync().ConfigureAwait(false);
-                try
+                using (await _logFileSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false))
                 {
                     using (var streamWriter = File.CreateText(_logFilePath))
                     {
@@ -188,11 +187,6 @@ namespace EventFlow.EventStores.Files
                         await streamWriter.WriteAsync(json).ConfigureAwait(false);
                     }
                 }
-                finally
-                {
-                    _logFileSemaphore.Release();
-                }
-
                 return committedDomainEvents;
             }
         }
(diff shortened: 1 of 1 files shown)


rasmus:
@t-johnson I have triggered a rebuild. The MSSQL tests seems more unstable than they use to 😢

Source/EventFlow/EventStores/Files/FilesEventPersistence.cs
```diff
@@ -167,20 +168,29 @@ public class EventStoreLog
                     committedDomainEvents.Add(fileEventData);
                 }
 
-                using (var streamWriter = File.CreateText(_logFilePath))
+                //only allow a single write access to the _logFilePath file at one time
+                await _logFileSemaphore.WaitAsync().ConfigureAwait(false);
```
rasmus: There's an `AsyncLock` in the EventFlow code base that supports cancellation and the `using` pattern, please use that.

t-johnson:
Thanks for the comment, I learned about AsyncLock now!

This whole function is already wrapped in an AsyncLock using statement... so perhaps the issue is not real?

t-johnson:
I modified according to your feedback.  the existing AsyncLock is not on a static member, so it will only lock for the current instance (fine for writing individual event files), while the new AsyncLock around the log file is static, to ensure only one instance gets at this one file.

the AppVeyor build seems stuck at the 'running tests' stage, but i was able to run all tests (minus the elastic search) on my local: Run test finished: 501 run (0:02:49.6750868).  I see others have had problems with AppVeyor and the sql instance, so ..  odd?

Source/EventFlow/EventStores/Files/FilesEventPersistence.cs
```diff
@@ -40,6 +40,7 @@ public class FilesEventPersistence : IEventPersistence
         private readonly IFilesEventLocator _filesEventLocator;
         private readonly AsyncLock _asyncLock = new AsyncLock();
         private readonly string _logFilePath;
+        private static AsyncLock _logFileSemaphore = new AsyncLock();
```
rasmus: EventFlow only uses static field for stuff that could be shared between different domains as EventFlow is design to be used as multiple instances within the same process.

rasmus:
☝️ It can be found here https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow/Extensions/EventFlowOptionsEventStoresExtensions.cs#L56

Source/EventFlow/EventStores/Files/FilesEventPersistence.cs
```diff
@@ -167,22 +168,25 @@ public class EventStoreLog
                     committedDomainEvents.Add(fileEventData);
                 }
 
-                using (var streamWriter = File.CreateText(_logFilePath))
+                //only allow a single write access to the _logFilePath file at one time
+                using (await _logFileSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false))
```
frankebersoll: I don't think we need an additional semaphore: The whole method is already being locked by `_asyncLock`, making an inner lock redundant.

frankebersoll:
@rasmus I tried changing the registration to singleton as you suggested, and added some unit tests. See #395

rasmus:
@t-johnson The PR from @frankebersoll is almost finished so I'm closing this as a duplicate

PR #394, opened 2017-10-21
Allow LocateSagaAsync to return null

frankebersoll:
I have a use case where a specific event might be part of a saga, but can also happen independently: Accounts can be created "on demand" during a payment process (`PaymentSaga`), but they might also be created beforehand.

Every time the `PaymentSagaLocator` encounters a standalone instance of the `AccountCreated` event, it won't find any payment metadata and cannot extract a payment id. So it returns a new id, leading to the saga's initialization - only to be discarded soon as this event doesn't start the saga. This seems counterintuitive.

I added the possibility to return `null` instead in order to short circuit the saga dispatcher process, as that was what I tried to do in the first place.

### DIFF
--- a/Source/EventFlow.Tests/IntegrationTests/Sagas/AlternativeSagaStoreTests.cs
+++ b/Source/EventFlow.Tests/IntegrationTests/Sagas/AlternativeSagaStoreTests.cs
@@ -21,6 +21,7 @@
 // IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 // CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using EventFlow.Aggregates;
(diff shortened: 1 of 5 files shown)


Source/EventFlow/Sagas/DispatchToSagas.cs
```diff
@@ -83,6 +83,13 @@ public class DispatchToSagas : IDispatchToSagas
             {
                 var locator = (ISagaLocator) _resolver.Resolve(details.SagaLocatorType);
                 var sagaId = await locator.LocateSagaAsync(domainEvent, cancellationToken).ConfigureAwait(false);
+
+                if (sagaId == null)
+                {
+                    _log.Verbose(() => $"Saga locator '{details.SagaLocatorType.Name}' returned null");
```
rasmus: Instead of `'{details.SagaLocatorType.Name}'`, write `'{details.SagaLocatorType.PrettyPrint()}'` to get a more readable type name for logging

Source/EventFlow.Tests/IntegrationTests/Sagas/AlternativeSagaStoreTests.cs
```diff
@@ -101,5 +102,15 @@ public void TearDown()
...
         }
+
+        [Test]
+        public void SagaLocatorReturningNullDoesntThrow()
+        {
+            // Arrange
+            var aggregateId = AlternativeSagaStoreTestClasses.SagaTestAggregateId.With(Guid.Empty);
+
+            // Act
+            _commandBus.Publish(new AlternativeSagaStoreTestClasses.SagaTestBCommand(aggregateId), CancellationToken.None);
```
rasmus: Is it possible to do an assertion, e.g. is the saga loaded?

frankebersoll: I updated the InMemorySagaStore to be verifiable.

rasmus:
Perfect, is this something you would like to see release asap?

frankebersoll:
This is not a critical issue, since there is an easy workaround: I just return a new saga id.

PR #395, opened 2017-10-22
FilesEventPersistence as singleton, added concurrency tests

### LINKED ISSUE #356: FilesEventPersistence.CommitEventsAsync causing IOException to occur
I'm noticing this intermittently, getting a `IOException` thrown when trying to commit the events.

I think what's possibly happening is that the commit is being called from multiple threads (perhaps because of multiple requests firing events at the same time?), and end up in a race condition at [FilesEventPersistence.cs#L170](https://github.com/eventflow/EventFlow/blob/fa315bc60362df21d1f5f474c8234b268122d66b/Source/EventFlow/EventStores/Files/FilesEventPersistence.cs#L170), where one thread has claimed access to the file in it's call to `File.CreateText`, while a second thread hits the same bit of code at the same time.

I can't reliably replicate this, just seems to happen spuriously when I'm firing simultaneous requests to my API.

### LINKED ISSUE #357: File store should be thread safe

frankebersoll:
I did some research with regard to issue #357 (File store should be thread safe):

- Added some unit tests that simulate concurrent access. Could _not_ reproduce any issues when accessing a single instance from multiple threads.
- However, added locking for public methods that didn't have it already.
- Made the `FileEventsPersistence` registration a singleton instance.

Fixes #356
Fixes #357

### DIFF
--- a/Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
+++ b/Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
@@ -22,6 +22,7 @@
 // CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
@@ -45,9 +46,12 @@ namespace EventFlow.Tests.UnitTests.EventStores
     [Category(Categories.Unit)]
     public class ConcurrentFilesEventPersistanceTests
     {
-        private const int DegreeOfParallelism = 10;
+        // Higher values have exponential effect on duration
+        // due to OptimsticConcurrency and retry
+        private const int DegreeOfParallelism = 15;
         private const int NumberOfEventsPerBatch = 10;
 
+        // All threads operate on same thingy
         private static readonly ThingyId ThingyId = ThingyId.New;
 
         private string _storeRootPath;
@@ -86,7 +90,7 @@ namespace EventFlow.Tests.UnitTests.EventStores
             var tasks = RunInParallel(async i =>
             {
                 var persistence = CreatePersistence("SameForAll");
-                await CommitEvents(persistence).ConfigureAwait(false);
+                await CommitEventsAsync(persistence).ConfigureAwait(false);
             });
 
             // Act
@@ -103,7 +107,7 @@ namespace EventFlow.Tests.UnitTests.EventStores
             var tasks = RunInParallel(async i =>
             {
                 var persistence = CreatePersistence(i.ToString());
-                await CommitEvents(persistence).ConfigureAwait(false);
+                await CommitEventsAsync(persistence).ConfigureAwait(false);
             });
 
             // Act
@@ -120,7 +124,7 @@ namespace EventFlow.Tests.UnitTests.EventStores
             var persistence = CreatePersistence();
             var tasks = RunInParallel(async i =>
             {
-                await CommitEvents(persistence).ConfigureAwait(false);
+                await CommitEventsAsync(persistence).ConfigureAwait(false);
             });
 
             // Act
@@ -145,15 +149,15 @@ namespace EventFlow.Tests.UnitTests.EventStores
             return new FilesEventPersistence(log, serializer, config, locator);
         }
 
-        private async Task CommitEvents(FilesEventPersistence persistence)
+        private async Task CommitEventsAsync(FilesEventPersistence persistence)
         {
             var events = Enumerable.Range(0, NumberOfEventsPerBatch)
                 .Select(i => new ThingyPingEvent(PingId.New))
-                .ToArray();
+                .ToList();
 
-            await Retry(async () =>
+            await RetryAsync(async () =>
             {
-                var version = await GetVersion(persistence).ConfigureAwait(false);
+                var version = await GetVersionAsync(persistence).ConfigureAwait(false);
 
                 var serializedEvents = from aggregateEvent in events
                     let metadata = new Metadata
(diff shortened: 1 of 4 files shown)


Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        private FilesEventPersistence CreatePersistence(string storePath = "")
+        {
+            var log = new NullLog();
+            var serializer = new JsonSerializer();
+            var config = ConfigurePath(storePath);
+            var locator = new FilesEventLocator(config);
+            return new FilesEventPersistence(log, serializer, config, locator);
+        }
+
+        private async Task CommitEvents(FilesEventPersistence persistence)
```
rasmus: Rename to `CommitEventsAsync`

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
```diff
@@ -0,0 +1,211 @@
...
+                var readOnlyEvents = new ReadOnlyCollection<SerializedEvent>(serializedEvents.ToList());
+
+                await persistence
+                    .CommitEventsAsync(ThingyId, readOnlyEvents, CancellationToken.None)
+                    .ConfigureAwait(false);
+            })
+            .ConfigureAwait(false);
+        }
+
+        private static async Task<int> GetVersion(FilesEventPersistence persistence)
```
rasmus: Rename to `GetVersionAsync`

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            var tasks = Enumerable.Range(1, DegreeOfParallelism)
+                .AsParallel()
+                .WithDegreeOfParallelism(DegreeOfParallelism)
+                .Select(action);
+
+            return tasks.ToArray();
+        }
+
+        private static async Task Retry(Func<Task> action)
```
rasmus: Rename to `RetryAsync`

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
```diff
@@ -0,0 +1,211 @@
...
+using EventFlow.TestHelpers.Aggregates.ValueObjects;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace EventFlow.Tests.UnitTests.EventStores
+{
+    [Category(Categories.Unit)]
+    public class ConcurrentFilesEventPersistanceTests
+    {
+        private const int DegreeOfParallelism = 10;
```
rasmus: Maybe bump these numbers a bit, the build servers are shared VMs and concurrency problems tends to be a bit hard to test on slow machines.

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            var config = ConfigurePath(storePath);
+            var locator = new FilesEventLocator(config);
+            return new FilesEventPersistence(log, serializer, config, locator);
+        }
+
+        private async Task CommitEvents(FilesEventPersistence persistence)
+        {
+            var events = Enumerable.Range(0, NumberOfEventsPerBatch)
+                .Select(i => new ThingyPingEvent(PingId.New))
+                .ToArray();
```
rasmus: Use `ToList()`

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
frankebersoll: Those tests are indeterministic by nature. I cranked it up to 15 while leaving the batch size at 10. I won't call these "sensibly tuned", but i got 100 consecutive green runs with those. If it fails on the build server somewhen, just tell me and I remove the tests. Or we can abstract the file system to mock long running tasks.

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
frankebersoll: Can you commit your .dotsettings? Then I could write a rule that prevents this in the future.

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
frankebersoll: TIL. I was using ToArray all the time as it intuitively seemed more lightweight. Now I will switch from array return types to IEnumerable and use ToList. Will take some time getting used to it.

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
rasmus: I don't have them as settings. Might set automatic inspections up instead. Its the general Microsoft async/await guideline

Source/EventFlow.Tests/UnitTests/EventStores/ConcurrentFilesEventPersistanceTests.cs
rasmus: Same for me 😄

rasmus: In EventFlow i use `IEnumerable` for returns that are lazy evaluated and `IReadOnlyCollection` for stuff that's collections and already evaluated

PR #400, opened 2017-10-25
Fixed #399: Made static Sha256 instance transient

### LINKED ISSUE #399: StringExtensions.ToSha256 is not thread-safe
I got bitten by a multi threading issue where multiple threads (xUnit tests) used `string.ToSha256()` at the same time. Obviously, Sha256Managed keeps some state during hash calculation, which isn't supposed to be shared. We should create instances on demand and dispose them afterwards.

frankebersoll:
StringExtensions.ToSha256() could not be used by concurrent threads because a static instance was involved. Fixed this issue by making it transient and disposing it.


rasmus:
@frankebersoll Thanks! I'll fix the build now and get this released asap

PR #402, opened 2017-10-26
Multiple aggregate ordering issue

houseme-brandon:
As per the commit comment:
Aggregate A has events up to V20
Aggregate B has events up to V1

If aggregate B started at V18 of aggregate A then it should be updated in that order when rebuilding the event store. However it was being ordered as such:
A V1
B V1
A V2
A V3
.
.
.
A V20


rasmus:
@houseme-brandon The CLA is merely to verify that your changes conform with the MIT license. I work at eBay so need the paperwork to be in order, even for my spare time projects

rasmus:
@houseme-brandon Could you write a test and update the release notes? then it would be perfect

houseme-brandon:
Sure I will get on that :)

PR #407, opened 2017-11-03
WIP: State Machine

frankebersoll:
This is the very first conceptual draft for a **state machine**, requesting for comments.

## Why?

For aggregates that use more than a few events and constantly need to check on some enum value if they are in the right, well, state.

## Definition syntax

```csharp
return builder
    .StartWith<WaitingForCoinsState>()
    .In<WaitingForCoinsState>()
        .When<CoinInserted>().Use<AddCoinTransition>()
        .When<Selected>().Use<CheckEnoughCoinsTransition>()
    .In<SelectedState>()
        .When<CoinInserted>().Ignore()
        .When<Selected>().Use((state, signal) => state.With(signal.Selection))
    .Build();
```
## Transition syntax

```csharp
public class CheckEnoughCoinsTransition : ITransition<WaitingForCoinsState, Selected>
{
    public IState Execute(WaitingForCoinsState currentState, Selected signal)
    {
        var change = currentState.Value - signal.Price;
        if (change >= 0)
            return new SelectedState(signal.Selection, change);

        return currentState;
    }
}
```

## Usage

```csharp
public void InsertCoin(int value)
{
    Emit(new CoinInserted(value));
}

public void Select(Selection selection, int price)
{
    Emit(new Selected(selection, price));
}
```

Emitting an event will look for a transition that matches current state and event type. If there is no transition, Emit fails. If executing the transition returns null or throws an exception, Emit also fails. If emit doesn't fail, it will continue as usual, finally applying the event. Applying does the same thing as above, but this time it will replace the current state with the new state returned from the transition.

## Issues

1. Emitting events before inspecting the state goes against how EventFlow works otherwise:
    - Events should only be created if something is allowed to happen, i.e. if there is a transition and the transition doesn't throw.
    - Should events and signals be different things? If so, how would a transition work?

        ``` csharp
        var transition = GetTransition(signal, CurrentState);
        var event = transition.Handle(signal, CurrentState); // Check inputs and current state
        if (event == null) throw;
        CurrentState = transition.Apply(event, CurrentState); // Execute transition
        ```

    - Could we then use commands as signals, and handle them implicitly? Dependency injection in transitions would make them comparable to command handlers. Maybe we don't need both.
1. Can we somehow enforce that states are immutable? Transitions should always return new instances instead of mutating the current state.
1. Should the state machine work at the aggregate level (which would need multiple implementations for `AggregateRoot` and `AggregateSaga`) or at the AggregateState level, pluggable into aggregates and sagas? How would that work?

### DIFF
--- a/Source/EventFlow.Tests/Exploration/StateMachine/Framework/IState.cs
+++ b/Source/EventFlow.Tests/Exploration/StateMachine/Framework/IState.cs
@@ -1,0 +1,6 @@
+namespace EventFlow.Tests.Exploration.StateMachine.Framework
+{
+    public interface IState
+    {
+    }
+}
--- a/Source/EventFlow.Tests/Exploration/StateMachine/Framework/IStateMachineDefinition.cs
+++ b/Source/EventFlow.Tests/Exploration/StateMachine/Framework/IStateMachineDefinition.cs
@@ -1,0 +1,12 @@
+using System;
+using EventFlow.Tests.Exploration.StateMachine.Framework.Transitions;
+
+namespace EventFlow.Tests.Exploration.StateMachine.Framework
+{
+    public interface IStateMachineDefinition
+    {
+        Type InitialStateType { get; }
+
+        ITransition GetTransition(TransitionKey key);
+    }
+}
--- a/Source/EventFlow.Tests/Exploration/StateMachine/Framework/ITransition.cs
+++ b/Source/EventFlow.Tests/Exploration/StateMachine/Framework/ITransition.cs
@@ -1,0 +1,14 @@
+using EventFlow.Aggregates;
+
+namespace EventFlow.Tests.Exploration.StateMachine.Framework
+{
+    public interface ITransition
+    {
+        IState Execute(IState currentState, IAggregateEvent signal);
+    }
+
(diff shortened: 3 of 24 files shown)


rasmus:
Maybe have a look at https://github.com/dotnet-state-machine/stateless

frankebersoll:
@rasmus I looked at different state machine implementations. Stateless seems to be the most common, but I couldn't find a way to use anything but enums or strings with it. What I want to do is combine a state ("In Progress") with data that belongs to this state ("Assigned to"). This would somehow replace how state is held for an aggregate. Please take a look at the example in my commit:

`WaitingForCoinsState` contains the current value of coins inserted, and nothing else, because only the value is needed. `SelectedState` contains how much change there is to return and which selection the user has made. 

Making only the relevant data available in the respective states leads to a workflow like this: `Current State + Event -> Transition -> New State` where state not only contains an enum value, but the actual aggregate state. Main benefits would be 1. keeping state that belongs together at one place (state classes) and 2. replacing Specs that check on the current state by a declarative approach (transition configuration).

If you don't see any value in it, I can continue this in my own repository and try to make a full example.

PR #411, opened 2017-11-19
Upgrade to net standard 2.0

JackHerring:
@rasmus it was only possible to target `netstandard 2.0` in the `MsSql` & `Sql `projects by using the latest, beta, version of `DbUp`. I can't see any other projects using pre-release nuget packages so I don't know if you'll be OK with that..?

Whilst I was there I upgraded the projects targeting `netstandard 1.6` to `2.0`, too. There are no code changes here, just `csproj `and removal of `#if` directives.

### DIFF
--- a/Source/EventFlow.MsSql/EventFlow.MsSql.csproj
+++ b/Source/EventFlow.MsSql/EventFlow.MsSql.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../Common.props" />
   <PropertyGroup>
-    <TargetFramework>net451</TargetFramework>
+    <TargetFrameworks>net451;netstandard2.0</TargetFrameworks>
     <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
     <GenerateAssemblyInfo>True</GenerateAssemblyInfo>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
@@ -30,5 +30,9 @@
     <EmbeddedResource Include="EventStores\Scripts\0001 - Create table EventFlow.sql" />
     <EmbeddedResource Include="EventStores\Scripts\0002 - Create eventdatamodel_list_type.sql" />
     <EmbeddedResource Include="SnapshotStores\Scripts\0001 - Create EventFlowSnapshots.sql" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.ComponentModel.Annotations" Version="4.4.0" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.4.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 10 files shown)


JackHerring:
@rasmus I have made the `DbUp `reference conditional on the targeted framework so that the `net451 `version doesn't need to use the pre-release nuget package. I wasn't sure if by '_Do you need to upgrade all the references to netstandard 2.0 as well?_' you meant did I need to re-target _other projects_ to `netstandard 2.0` or having retargeted them did I need to update the nuget packages. I have assumed the latter, and made that change but if you meant the former then no, it wasn't necessary I just figured it was possible so why not :) But will revert if you would like.

Source/EventFlow.MsSql/EventFlow.MsSql.csproj
```diff
@@ -31,4 +31,8 @@
     <EmbeddedResource Include="EventStores\Scripts\0002 - Create eventdatamodel_list_type.sql" />
     <EmbeddedResource Include="SnapshotStores\Scripts\0001 - Create EventFlowSnapshots.sql" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.ComponentModel.Annotations" Version="4.4.0" />
```
rasmus: aren't these two package references required for netstandard? please add a condition

Source/EventFlow.Sql/EventFlow.Sql.csproj
```diff
@@ -23,15 +23,20 @@
...
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="dbup" Version="4.0.0-beta0002" />
+  </ItemGroup>
+  
   <ItemGroup>
     <PackageReference Include="Dapper" Version="1.50.2" />
-    <PackageReference Condition="'$(TargetFramework)' == 'net451'" Include="dbup" Version="3.3.5" />
-    <PackageReference Condition="'$(TargetFramework)' == 'netstandard1.6'" Include="System.ComponentModel.Annotations" Version="4.3.0" />
+    <PackageReference Include="System.ComponentModel.Annotations" Version="4.4.0" />
```
rasmus: please re-add the condition

Source/EventFlow.Sql/EventFlow.Sql.csproj
```diff
@@ -23,15 +23,20 @@
   <ItemGroup>
     <Folder Include="Properties\" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
+    <PackageReference Include="dbup" Version="3.3.5" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="dbup" Version="4.0.0-beta0002" />
```
rasmus: image

From https://www.nuget.org/packages/dbup/4.0.0-beta0002

PR #421, opened 2018-01-25
ReflectionHelper.CompileMethodInvocation

domenkogler:
http://docs.geteventflow.net/Aggregates.html#events states that Appy method _can_ be `private`

 - recognise private methods
 - update test


rasmus:
@domenkogler Thanks for the fix!. Could you update the release notes as well. I'll merge when done.

domenkogler:
@rasmus Release notes updated.

PR #428, opened 2018-02-22
WIP: Less breaking implementation of LibLog

rasmus:


rasmus:
@wgtmpeters I did another implementation that doesn't change the `ILog` interface. Would this suite your needs?

rasmus:
Obsoletes #258

wgtmpeters:
I will look at it soon, we are looking at making it possible to use serilog to do structured logging. We have used decorators that uses our logging for parts we wanted to log serilog. I think what you are doing now will make it possible for us to use your use serilog for all logging calls of eventflow right? But that will not give us structured logging right?

rasmus:
@wgtmpeters No, it will not. As not all log frameworks supports structured logging EventFlow is left with simple messages. We have been using Serilog at work for several years and it might be interesting to make EventFlow logging a bit more intelligent to allow structured logging

rasmus:
Released in 0.54.3261

PR #445, opened 2018-03-29
Support for Microsoft.Extensions.DependencyInjection: EventFlow.ServiceProvider

frankebersoll:
Implements an `IServiceRegistration` and `IResolver` adapter for **Microsoft.Extensions.DependencyInjection**.

Pain points:

- `ServiceResolver` doesn't support any introspection or extensibility points, so there is no possibility to easily attach the decorator service to activated instances. Instead, we always register factory methods and wrap the create instances using the decorator.
- In ASP.NET Core, any initialization logic that actually *does something* (besides registering services) should not happen in `ConfigureServices`, so we cannot call `CreateResolver` there. In order to run the bootstrappers, we use the [`IHostedService`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostedservice?view=aspnetcore-2.0) interface instead which is called by the host during startup.


Source/EventFlow.ServiceProvider/Registrations/ServiceCollectionServiceRegistration.cs
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+
+        /// <summary>
+        ///     Ensures that the <see cref="Bootstrapper" /> is run in an ASP.NET Core
+        ///     environment when EventFlow is configured into an existing ServiceCollection
+        ///     instance and <see cref="CreateResolver" /> is not used.
+        /// </summary>
+        // ReSharper disable once ClassNeverInstantiated.Local
+        private class HostedBootstrapper : IHostedService
```
rasmus: @frankebersoll If anyone simply pulls this NuGet into a .NET Core console application, they won't know why their code isn't working as there's nothing to start the hosted service.

If this is used in a ASP.NET Core application, then the order of which the `StartAsync` of all hosted services determine if its valid to start using EventFlow. It might be better to simply make the dependency  on ASP.NET Core and use `IApplicationLifetime.ApplicationStarted` instead. It seems its [invoked before any hosted services](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/hosted-services#ihostedservice-interface).

rasmus: Might be better to remove and move the dependency to the ASP.NET Core package instead

rasmus:
Hi @frankebersoll, I have been struggling with a good solution but it all boils down to developers wanting this and the functionality isn't there.

So, how about creating a read-me file that notifies developers that they need to execute all `IBootstrap` implementations before they can start using EventFlow?

frankebersoll:
I will try that by the of the week. There is another issue that needs investigation: A coworker got an exception when he tried to inject an IEventStore into a service used by a controller. I think it was related to scopes 😐 Will talk to him about it later today.

rasmus:
Thanks

frankebersoll:
## Bootstrapping
- Added `IBootstrapper` that encapsulates ordering and running the IBootstrap instances, so it can be reused across all registrations and easily triggered from the outside.
- Registered `IBootstrapper` default implementation in all registrations as singleton, so it can track if it has already been run.
- Used `IBootstrapper` instead of the existing logic in `CreateResolver` implementations.
- Used `IBootstrapper` from `IHostedService` implementation (ASP.NET Core) or `AutofacStartable`.

## Issue with scoped instances
I added some tests to reproduce an issue where using a scoped instance from an `IQueryHandler` crashes: ASP.NET Core uses scopes for each request and `AddDbContext<TContext>()` registers the Entity Framework context as scoped instance, so it will exist once per request.
Because `IQueryProcessor` is a singleton by default, it uses the root `IResolver` instead of a scoped one. It even caches handler instances, so we can't really use scoped dependencies for query handlers for the time being.
While the ServiceProvider container does some checks for scoped access and throws whenever a scoped object tree trys to access the root resolver, Autofac does not, which might result in some hard to debug instance problems.

I think we need to sort this out before we publish the package, what do you think?

Source/EventFlow.ServiceProvider/EventFlow.ServiceProvider.csproj
```diff
@@ -0,0 +1,33 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <GenerateAssemblyInfo>True</GenerateAssemblyInfo>
+    <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
+    <PackageRequireLicenseAcceptance>True</PackageRequireLicenseAcceptance>
+    <Title>EventFlow.ServiceProvider</Title>
+    <Authors>Rasmus Mikkelsen</Authors>
```
rasmus: @frankebersoll Please put your name in the `Authors`. I didn't write it and I don't deserve the credits.

rasmus:
... and as soon as the scope problem is fixed as well

frankebersoll:
Thank you! 

I don't mind who is written in the file headers. I think it's wrong as it is now, because legally I am the copyright owner as I have written the code (in my off time, otherwise it would be my employer or our client) and we don't have some kind of Copyright Transfer Agreement. Obviously, I also intend to publish it under MIT license, so it will only matter if somebody wants to use it without the context of MIT, in which case they needed to contact me. Wouldn't it be easier to add a Contributors.md?

What do you mean by recommendation? What would you recommend?

You can merge and release it as you please, I moved the scope tests to another branch.

rasmus:
Its only recently that I though about the contributions that are added to EventFlow by others but in files containing only my name. I would very much like to acknowledge the work others put in EventFlow as well but I'm unsure how the format should be.

As you said the file headers is just one of the issues.

Adding the `Contributors.md` is an start.

As for the license I think the (d) clause of the [CLA](https://cla-assistant.io/eventflow/EventFlow?pullRequest=445) covers it. I will have to contact our legal department to be sure though. eBay isn't directly involved in EventFlow, but does offer legal support as the project is used by eBay and I'm an employee.

rasmus:
Recommendation as in job recommendation if you ever apply for a job at e.g. eBay 😉

rasmus:
Might be that `DependencyInjection` is the right name. Will have a look at some reference projects and then rename projects before next release

Issue #294, opened 2017-01-08
Import a large number of events


minhhungit:
Hi team,

My team is considering to use Eventflow for our project.
- Problem is we need to import large events to SqlServer Event Store, so we're writing a console application to load data from existed tables and send command to build events.
- We always update large data into database (about 2-3 milion records by csv file) once a week.

Our database has a lot records, about **30 milions** records for all tables and it will get more in next years
I think the best way is disable `ReadModels` feature, we just need insert events into EventStore and replay them. But if we just insert them one by one then it will very slow.
I also think to use [table-value-param](http://stackoverflow.com/a/25773471/3007147) to insert batch event into EventStore table, table-value param is fastest way that I know.
- Can anyone show me how to implement it (**with table-value param**) or suggestion ? 
- Did we have any option that disable `ReadModel` feature?
- Should I design to use 1 table for 1 aggregrateRoot to decrease event version on each aggregateRoot?

Thank you!

rasmus:
I'm not at a computer right now, but for the table value you can use the built-in that EventFlow has implemented and look at the implementation from the EventFlow code base.

I'll see if I have any input to the rest when I'm at a computer.

rasmus:
@minhhungit Got computer access

Just wanna give you a few ideas, weather or not any of them are good in your situation depends on your application and your preference.

- You could make your own implementation of the `MsSqlEventPersistence` and use that on your existing event store, then you wouldn't need to migrate anything. You would need to implement the `EventJsonSerializer` as well. Don't know if its easy, but it might be something to consider as it will allow you to go back
- The table value type in EventFlow is called `TableParameter<>` and can be used directly via `IMsSqlConnection.InsertMultipleAsync<long, EventDataModel>(...)`. The SQL for the type is here https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow.MsSql/EventStores/Scripts/0002%20-%20Create%20eventdatamodel_list_type.sql and the wrapper that handles the SQL magic is called `TableParameter<>` in the EventFlow code base. Your console application could use that. Not sure about the performance though, haven't tested it.
- Disabling the read model feature is rather simple, just don't use it 😄 its opt-in. If you want to disable it further though, you could create an empty implementation of the `IDomainEventPublisher` and register that in the IoC container
[...]

Issue #191, opened 2016-01-24
Stream events to aggregates


rasmus:
Instead all events into memory first and applying them after, "steam" them instead. Either read small batches, or individual events (depending on the underlying persistence model) to minimize memory consumption. If possible, it should be configurable or even optional.

rasmus:
Will have a look at this with `IAsyncEnumerable`, maybe for 1.0

Issue #49, opened 2015-04-28
Archive event streams


rasmus:
As an alternative to the of deleting event streams described in #48, we should provide functionality to ease the process of archiving an aggregate event stream.

rasmus:
Consider replacing all events in the original store with a special archived event, then if the aggregate is loaded, throw a `AggregateArchivedException` or something similar. This would make it impossible to reuse the same aggregate ID twice.

rasmus:
@MartinHave What do you think about this idea? Specifically the concept of throwing an exception if the aggregate is loaded and archived.

MartinHave:
@rasmus Why would we want to do that? Could you specify when this would be handy? As a base rule, we should never do anything else than appends on the event store.

rasmus:
@MartinHave When a aggregate is deleted, then we would like to archive the event stream somewhere else to minimize storage so that our event store doesn't grow indefinitely. Thus we would like to be able to archive an aggregate, i.e., move the event stream to some cheaper storage. The idea was, as we move the event stream, then if a aggregate by accident is created with the same ID as an archived, then it should throw an error.

MartinHave:
@rasmus There are many cases where it is very important to keep the event stream intact. What about new views that needs to aggregate across many events even those that have ended there aggregate lifetime. I I would want to create a new statistics view, then I would not get a complete picture. I would argue that we should NEVER move event from the event stream.

rasmus:
@MartinHave How about creating a concept of a primary and secondary event stores? Then if you want to archive an infrequently used (or deleted) aggregate, then move the event stream to the secondary and cheaper event store. The event stream would still be accessible, but make take a bit longer to load. The application should load the aggregate normally and it shouldn't be possible to tell if you load an aggregate from primary or secondary on anything but the load time.

PiotrBrzezianski:
@MartinHave @rasmus This subject came up recently for us in context of GDPR and "right to be forgotten". Any thoughts on how append-only world should deal with such cases?

rasmus:
@PiotrBrzezianski its easy, delete everything. In cases where you can't delete, use a strong one-way hash (please check with your legal department in a case-by-case to verify that this indeed a possibility).

The "append-only" and "don't modify event streams" falls apart when in comes to GDPR. By law you are to delete everything. Even event times are considered PII as they can be used to identify individuals based on when they were online.

[...]

PR #464, opened 2018-05-14
PoC: Event streaming

rasmus:
### Relevant for ###

- Loading aggregates with a huge amount of events and no snapshots (#191)
- Importing events to the event persistence (#294)
- Exporting huge aggregates (#49)

Current implementation uses the `System.Interactive.Async` NuGet package, but C# 8 might introduce `IAsyncEnumerable` as part of the language.

### To do ###

- [ ] Determine why resources aren't released
- [ ] Remove reference to `System.Interactive.Async`
- [ ] Write more tests
- [ ] Rework read model re-populating to use streaming


rasmus:
Hi @wgtmpeters I just haven't done the implementation for EventStore yet, but I'll look into it. It should be relatively easy as the current implementation is already loading events in batches, but is currently putting them [into a list](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow.EventStores.EventStore/EventStoreEventPersistence.cs#L72) before returning.

I haven't looked into how bulk inserting would work for EventStore.

But I hope these changes, along with event archiving, will make it easier to manage huge event stores and aggregates.

rasmus:
Too old

PR #466, opened 2018-05-23
Change Primary Key name for MSSQL Snapshot Store

frankebersoll:
### Problem ###

Using MSSQL EventStore and SnapshotStore in the same database seems to cause conflicts:

```
System.Data.SqlClient.SqlException (0x80131904): There is already an object named 'PK_EventFlow' in the database.
Could not create constraint or index. See previous errors.
```

### Solution ###

Renaming the snapshot PK to `PK_EventFlowSnapshots` seems to do the trick.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,8 @@
 
 * Fix: Commands are now correctly published when no events are emitted from a saga
   after handling a domain event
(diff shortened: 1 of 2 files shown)


rasmus:
@frankebersoll That just shows that snapshots hasn't been used together yet. Interesting.

rasmus:
Will create a release now

PR #468, opened 2018-05-29
Ported EventStore to .NET standard

anthonyhuebers:
Ported EventStore to .NET standard


rasmus:
Found the problem why the optimistic concurrency exception wasn't thrown. It seems like EventStore won't throw wrong version exceptions if the events added have the same guid. Thus a new guid is required for every event and the `EventId` cannot be used as its deterministic based on the aggregate ID and the aggregate sequence number. I'll finish up PR #475

wgtmpeters:
Maybe it's easier to just make the changes for
Netstandard in your clean up branch instead of updating this one? . It's only in the csproj file.

rasmus:
Hi @wgtmpeters. I updated the branch and as you said, there's only changes to the project file. If you could update the release notes and sign the CLA then we can merge it in 😄

rasmus:
Will do in another PR as suggested

PR #475, opened 2018-06-06
EventStore cleanup

### LINKED ISSUE #469: Is EventStore support production ready?
Can I remove the "only for test (for now)" for EventStore in the README.md?

#### tl;dr

The current implementation of EventStore was something I made as a PoC and thus I haven't personally verified the implementation in a production environment.

This is a call-out to anyone either currently using the implementation or someone experienced in working with EventStore to verify that the implementation is "production ready".

rasmus:
Fixes #469

### DIFF
--- a/Source/EventFlow.EventStores.EventStore/EventStoreEventPersistence.cs
+++ b/Source/EventFlow.EventStores.EventStore/EventStoreEventPersistence.cs
@@ -122,7 +122,7 @@ namespace EventFlow.EventStores.EventStore
             var eventDatas = serializedEvents
                 .Select(e =>
                     {
-                        var guid = Guid.Parse(e.Metadata["guid"]);
+                        var guid = e.Metadata.EventId.GetGuid();
                         var eventType = string.Format("{0}.{1}.{2}", e.Metadata[MetadataKeys.AggregateName], e.Metadata.EventName, e.Metadata.EventVersion);
                         var data = Encoding.UTF8.GetBytes(e.SerializedData);
                         var meta = Encoding.UTF8.GetBytes(e.SerializedMetadata);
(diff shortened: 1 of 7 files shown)


Source/EventFlow.EventStores.EventStore/EventStoreEventPersistence.cs
```diff
@@ -122,7 +122,10 @@ private static Position ParsePosition(GlobalPosition globalPosition)
             var eventDatas = serializedEvents
                 .Select(e =>
                     {
-                        var guid = Guid.Parse(e.Metadata["guid"]);
+                        // While it might be tempting to use e.Metadata.EventId here, we can't
+                        // as EventStore won't detect optimistic concurrency exceptions then
+                        var guid = Guid.NewGuid();
```
volak: Do you have a way to make event ids deterministic here?  EventStore has the ability to disregard duplicate event ids - using a brand new guid every event loses the (albeit rarely needed) guarantee and event isnt written twice.

See https://eventstore.org/docs/dotnet-api/optimistic-concurrency-and-idempotence/index.html#idempotence

volak: What I do is use the message id's Guid and increment that per event raised.  If a command is run twice for some reason it generates a series of duplicate event ids

rasmus: The `e.Metadata.EventId` _is_ deterministic but as its currently only dependent on the aggregate ID and the event sequence number (and not the event data), there's no guarantee that it is indeed the same event. If it was possible to make 100% sure that developers implemented their events using value objects and these were implemented correctly, then it might be possible. However, if the command is executed twice its far simpler to make the executed command idempotent in the aggregate, e.g. resulting in no events if the same command is executed again. The aggregate will still be reloaded, but its easier to manage.

Agree?

volak: yeah if you have another way to know if the command was already run then that works great.  For me I just use the command's message id - if you have another way then :+1:

PR #476, opened 2018-06-06
PoC: MSSQL read model schema generator

rasmus:
Investigate if its possible to create a simple read model schema creation for MSSQL


rasmus:
Too old

PR #490, opened 2018-06-14
Release notes: Delete read models

frankebersoll:


rasmus:
I was too fast, didn't get enough sleep

rasmus:
thanks

PR #497, opened 2018-06-24
Rename project to align with Microsoft NuGet package

rasmus:
@frankebersoll the rename we discussed, what do you think?

The idea with the rename is make it easier for developers to choose the right package as there are quite a few EventFlow packages.


frankebersoll:
Yes, I like this better. Microsoft didn‘t make it easy by giving those things such generic names.

Issue #260, opened 2016-10-27
EntityFramework support


rasmus:
Provide NuGet package that enables MS Entity Framwork

Create another repo.

If anyone wants to be maintainer on such a package and get it hosted in the EventFlow org. Let me know.

pichardoSoftware:
I could help you with this, but I need more information.

rasmus:
@pichardoSoftware That would be great!

- The EF support is basically two parts, event persistance and read model support. I suggest you pick one and focus on that and get that PR finished first
- The project should be called EventFlow.EntityFramework and should contain both read model and event persistance support
- Have a look at the EventFlow.MsSql project source layout, it should be something like that
- Be sure to create integration tests. Its faily simple as there's a "suite" of tests that you can inherit from (https://github.com/eventflow/EventFlow/tree/develop/Source/EventFlow.TestHelpers/Suites). All the other projects use those for integration tests, so there should be some good examples
- Don't put any effort into updating build scrips and NuSpec files, as I hope to be able to close #257 soon and thereby making it much easier

If you have any questions, please ask. I preparing to move to a new house with the next three weeks, so I will not be able to provide that much in-depth support, but I'll try to answer any questions ASAP.

rasmus:
Forgot some basic stuff

- Event persistence, create implementation of `IEventPersistence`
- Read model support, implement `ReadModelStore<>`

You can have a look at the in-memory or MSSQL implementations for some general idea on how to do it

pichardoSoftware:
Ok, I'll check it out

rasmus:
@pichardoSoftware PR #257 is closed now and will make it easier to get started on new projects

rasmus:
@pichardoSoftware If you had problems working with EventFlow due to git lfs, give it another go as it has been removed

frankebersoll:
I have a working implementation for Entity Framework 6 persistence and read models.

- Would people use this (our company does)?
- EF6 / EF Core / Why not both?

wgtmpeters:
We are actually interested in this as well. Maybe you can get a branch up and running. If there is any help needed I might contribute as well.

rasmus:
I have received a few requests on it by mail as well

wgtmpeters:
@frankebersoll  Do you have any plans for setting up branch for this? We would really like to use this as well.

frankebersoll:
[...]

PR #500, opened 2018-06-26
WIP: Entity Framework

frankebersoll:
## Todo
- [x] Read Model
- [x] Snapshots
- [ ] Refactor configuration
- [ ] DbContext lifetime management
- [ ] Logging

## Nice to have
- [ ] Postgres integration test
- [ ] `byte[]` Version/Timestamp
- [ ] Value Types for Id properties
- [ ] Entity Framework 6

See also #260


Source/EventFlow.EntityFramework/Extensions/DbContextExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+
+namespace EventFlow.EntityFramework.Extensions
+{
+    public static class Bulk
```
frankebersoll: I'm not proud of this one. Please help. See comment for further information.

frankebersoll: There is no way to delete entities without querying them first: https://www.google.de/search?q=ef+core+"delete+where"

There seems to be a package that supports batch delete, but 1.) that's lots of code, 2.) it needs a reference to `Microsoft.EntityFrameworkCore.Relational` (that we don't need otherwise) - and 3.) obviously it needs to hardcode SQL commands for the major providers. Which means it only works for those, probably: https://github.com/zzzprojects/EntityFramework-Plus/blob/master/src/shared/Z.EF.Plus.BatchDelete.Shared/BatchDelete.cs

This affects both Read Stores and Snapshot Stores.

rasmus: Sorry, no help here, haven't worked with EF in almost five years. Pretty sure I'm outdated (we use Dapper for almost everything)

rasmus:
Implemented in #514

PR #505, opened 2018-06-30
ApplyEvents: Better support for snapshots

frankebersoll:
> - Aggregates must not assume that snapshots are created with increasing aggregate sequence numbers
>
> *Snapshot documentation*

I still managed to create a bug because of not having the actual Version value inside the `Apply` methods when snapshots were used.

This PR changes how events are applied in regard to aggregate sequence numbers:
- Possibly breaking: Removed `ApplyEvents(IEnumerable<IAggregateEvent>)`, which was only used by `AggregateRoot` itself and inlined it, in order to validate the current version for each applied event.
- Added `fromVersion` parameter: Sets the starting version after checking for initial state.
- Removed `domainEvents.Max(e => e.AggregateSequenceNumber)` because the Version should already be incremented

### DIFF
--- a/Source/EventFlow/Aggregates/AggregateRoot.cs
+++ b/Source/EventFlow/Aggregates/AggregateRoot.cs
@@ -135,22 +135,9 @@ namespace EventFlow.Aggregates
             return domainEvents;
         }
 
-        public void ApplyEvents(IReadOnlyCollection<IDomainEvent> domainEvents, int fromVersion = 0)
+        public void ApplyEvents(IReadOnlyCollection<IDomainEvent> domainEvents)
         {
             if (domainEvents == null) throw new ArgumentNullException(nameof(domainEvents));
-
-            if (Version > 0)
-            {
-                throw new InvalidOperationException($"Aggregate '{GetType().PrettyPrint()}' with ID '{Id}' already has events");
-            }
-
-            if (!domainEvents.Any())
-            {
-                Version = fromVersion;
-                return;
-            }
-
-            Version = fromVersion;
 
             foreach (var domainEvent in domainEvents)
             {
--- a/Source/EventFlow/Aggregates/AggregateRoot.cs
+++ b/Source/EventFlow/Aggregates/AggregateRoot.cs
@@ -135,46 +135,49 @@ namespace EventFlow.Aggregates
             return domainEvents;
         }
 
-        public void ApplyEvents(IReadOnlyCollection<IDomainEvent> domainEvents)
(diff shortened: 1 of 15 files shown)


frankebersoll:
This one needs a thorough review I think, as it affects the inner workings of event application, that happens all the time. Tests seem to pass, coding is finished so far.

Source/EventFlow/Aggregates/AggregateRoot.cs
```diff
@@ -135,46 +135,49 @@ protected virtual void Emit<TEvent>(TEvent aggregateEvent, IMetadata metadata =
             return domainEvents;
         }
 
-        public void ApplyEvents(IReadOnlyCollection<IDomainEvent> domainEvents)
+        public void ApplyEvents(IReadOnlyCollection<IDomainEvent> domainEvents, int fromVersion = 0)
```
rasmus: I don't think exposing this is a good idea. If its only for snapshots, then we should move setting the version to the `SnapshotAggregateRoot` and simply remove the `Version > 0` check.

Source/EventFlow/Snapshots/SnapshotAggregateRoot.cs
```diff
@@ -69,18 +69,14 @@ public abstract class SnapshotAggregateRoot<TAggregate, TIdentity, TSnapshot> :
 
             await LoadSnapshotContainerAsync(snapshot, cancellationToken).ConfigureAwait(false);
 
+            var version = snapshot.Metadata.AggregateSequenceNumber;
```
rasmus: By simply writing

```csharp
Version = snapshot.Metadata.AggregateSequenceNumber;
```

wouldn't that ensure that events versions are correct during apply after a snapshot load?

Source/EventFlow/Aggregates/AggregateRoot.cs
frankebersoll: You're right, we already have the sequential checks now for each event, so it will fail further down if you try applying event no. 1 to an aggregate that is not in Version 0, while it still supports setting the version beforehand. The additional check seems obsolete.

Source/EventFlow/Snapshots/SnapshotAggregateRoot.cs
frankebersoll: Yes. Fixed!

frankebersoll:
🎁

rasmus:
Very nice 😄

PR #517, opened 2018-08-22
Update System.Data.SQLite with netstandard2.0 target

coding4food:
New version of System.Data.SQLite adds `netstandard2.0` target, which enables to use `EventFlow.SQLite` in `netcoreapp` projects.


rasmus:
@coding4food Sorry for the delay. Its release now in 0.64.3598

PR #520, opened 2018-08-27
Retry unstable tests

rasmus:
😢 I wish these were stable but as they are now they are inconvenient  when developing something else for EventFlow.


rasmus:
Update fluent assertions separately

Issue #524, opened 2018-08-30
Exception when using EventStores.EventStore hosted in aspnet.core (linux) on docker 


TorMartin74:
```
11:28:13:799 [Debug]: Execution of command 'CreateConfigurationCommand' with ID 'command-0e9384ac-804f-4d94-9075-fe7e8490e910' on aggregate 'ConfigurationDomain.ConfigurationAggregate' failed due to exception 'ObjectDisposedException' with message: Cannot access a disposed object.
Object name: 'EventFlow v0.64.3598.0'. - System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'EventFlow v0.64.3598.0'.
   at EventFlow.EventStores.EventStore.EventStoreEventPersistence.LoadCommittedEventsAsync(IIdentity id, Int32 fromEventSequenceNumber, CancellationToken cancellationToken)
   at EventFlow.EventStores.EventStoreBase.LoadEventsAsync[TAggregate,TIdentity](TIdentity id, Int32 fromEventSequenceNumber, CancellationToken cancellationToken)
   at EventFlow.Aggregates.AggregateRoot`2.LoadAsync(IEventStore eventStore, ISnapshotStore snapshotStore, CancellationToken cancellationToken)
   at EventFlow.Aggregates.AggregateStore.LoadAsync[TAggregate,TIdentity](TIdentity id, CancellationToken cancellationToken)
   at EventFlow.Aggregates.AggregateStore.<>c__DisplayClass10_0`3.<<UpdateAsync>b__0>d.MoveNext()
```

rasmus:
@TorMartin74 I'm having a hard time reproducing the problem. Could you share some details about which object that throws the exception.

Have a look at the test in #525 here https://github.com/eventflow/EventFlow/pull/525/files#diff-a960ceab79a6fff3052f97f2415e1303R300

wernert72:
Not sure if it is related but i have the same issue only in [UpdateAsync](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow/Aggregates/AggregateStore.cs).
It happends when trying to resolve IDomainEventPublisher in my case.
Maybe [this so thread](https://stackoverflow.com/questions/51169733/net-core-2-1-objectdisposedexception-cannot-access-a-disposed-object-object) is related ?

I'm using DependencyInjection Pckg running latest version of .net core.
In my case no docker and on windows.

TorMartin74:
Some details:
[...]

PR #525, opened 2018-08-30
WIP: EventStore disposed

rasmus:


rasmus:
@frankebersoll I'm having a look at #524 and in doing so I cleaned up the internal test IoC and started to use the context in existing tests. I renamed the `IDbContext` to `IScopedContext` to better reflect what it is.

Let me know if you have any input.

PR #425, opened 2018-02-07
WIP: Use Docker container to run RabbitMQ integration tests

rasmus:
Instead of using an external RabbitMQ integration server, use a Docker container running locally.

This could also be used for e.g. PostgreSQL integration tests.

PR #528, opened 2018-08-31
Changes to use RabbitMQ docker container

idubnori:
Description
====
 - Run RabbitMQ container before running build script by `before_build:` section
    - An different approach from the #425 

Notes
====
 - Increased about 4 minutes of build time by thick docker image (5GB) downloading time
    - Solutions might be any of followings
         - Use a linux image (< 100MB), if we could get [a premium subscription](https://help.appveyor.com/discussions/questions/2652-windows-server-2016#comment_45617459)
         -  Make or find out more thin container image
         -  Caching docker image by AppVeyor setting (I'm not sure the details)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,15 +5,21 @@ version: 0.65.{build}
 
 skip_tags: true
 
+before_build:
+  - ps: |
+      docker pull $env:RABBITMQ_DOCKER_IMAGE_NAME
+      $cid = docker run -d $env:RABBITMQ_DOCKER_IMAGE_NAME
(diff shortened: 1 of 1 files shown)


idubnori:
@rasmus Elasticsearch tests failed. Could you please trigger again?
Have checked CI test successful on my forked side.

rasmus:
@idubnori I triggered the build again.

The addition to the build time would be a problem if I could expect builds to start right away, but as it is now with the free plan there's no guarantee. I think this is a step in the right direction as it removes the dependency on the external provider. I'll leave the other PR open as it could be interesting to see when Windows Server 2019 is released, which should have native support for Linux containers. This should hopefully allow us to run Linux containers on the free plan.

PR #532, opened 2018-09-04
Fix: Stack overflow in ValidateRegistrations

### LINKED ISSUE #523: ResolverExtensions.ValidateRegistrations throws StackOverflowException
Let's say we have a command handler decorator like this:
```
public class SomeCommandHandlerDecorator<TAggregate, TIdentity, TResult, TCommand> :
        CommandHandler<TAggregate, TIdentity, TResult, TCommand>
        where TAggregate : IAggregateRoot<TIdentity>
        where TIdentity : IIdentity
        where TResult: IExecutionResult
        where TCommand : ICommand<TAggregate, TIdentity, TResult>
    {
        public SomeCommandHandlerDecorator(ICommandHandler<TAggregate, TIdentity, TResult, TCommand> inner)
        {
            ...
        }

        ...
    }
```
in the same assembly as other command handlers.

When we register EventFlow types with `EventFlowOptionsDefaultExtensions.AddDefaults` (which calls `EventFlowOptionsCommandHandlerExtensions.AddCommandHandlers`) this open generic type gets registered as command handler. Any attempts to resolve it cause `StackOverflowException` (registration validation does that). Probably, some additional checks are needed to register only bound generic types as service implementations in built-in container.

frankebersoll:
Added checks in every Add* method for components that could potentially be decorated to filter out any types that have a constructor that takes a dependency on the same interface. This should catch any cases where a decorator gets incorrectly resolved, which leads to a stack overflow exception during resolution.

Added a failing to reproduce the issue.

This fixes #523

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,9 @@
 ### New in 0.64.3598 (released 2018-08-24)
 
 * New: Added .NET standard support for SQLite
+* Minor: Fixed stack overflow in `ValidateRegistrations` when decorator
+  components are co-located together with other components that are registed using
+  `Add*`-methods
 
 ### New in 0.63.3581 (released 2018-08-07)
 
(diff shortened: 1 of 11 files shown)


frankebersoll:
Branches got all mixed up. Game over.

PR #537, opened 2018-09-19
Fix: Performance issue in EventFlow.PostgreSql

idubnori:
- Changed the insert sql to use `RETURNING` clause in commit events
    -  Especially, improved the performance in high load, based on SQL cost analysis result and integratational performance measurement


idubnori:
@rasmus Ready to merge.

PR #538, opened 2018-09-21
Updated release notes for EntityFramework

idubnori:


idubnori:
@rasmus Missed it in #514, Sorry.

PR #545, opened 2018-10-12
Registering postgresql read model stores with their interfaces instead of their class types. Fixes exceptions on circular component dependencies when using autofac as IoC container.

identify1980:
When using the PostgreSql integration alongside with Autofac intgration, exceptions are thrown for circular component dependencies. As a fix the service registrations for read model stores will be done by using their interfaces instead of their class types.


rasmus:
I have a few fixes that I would like to get time for this weekend, I'll bundle this fix together with that, but release before Monday regardless

PR #552, opened 2018-10-16
Read model definition service

rasmus:
Experiment with a read model definition service

- Easy access to a list of known read model types
- Potential upgrade of outdated read model to allow run-time migration of old data


rasmus:
Too old

PR #559, opened 2018-10-24
WIP: migrate mongodb from EventFlow.MongoDb repo

wgtmpeters:
- [x] Migrate sources from repo
- [X] Fix EventStore tests (2 failing currently)
- [x] Find better place to create / ensure the event index exists
- [x] Use TransientFaultHandler wrapper around the exception handling of the optimisticconcurrency implementation
- [x] Add/Fix ReadStore tests, only one failing for the deletion of readmodels  test.
- [X] Add SnapshotStore tests


wgtmpeters:
@rasmus started working on migrating the MongoDB sources, some tests are failing for the event store part that I will look into.
But I was also adding tests for the ReadStore integration part but cannot really figure out why it is failing.
I keep getting exception in the ValidateRegistrations, I think because it cannot find the registration for the injected service in the queryhandlers:
       public MongoDbThingyGetQueryHandler(
            MongoDbReadModelStore<MongoDbThingyReadModel> mongeReadStore)

The same setup I have been using in my own solution and that worked without any validation exceptions from the resolver.

You have any idea where to look?

wgtmpeters:
@rasmus I have all integrations tests running now, but there is some extra code in the repo that is not tested.
It is about the IInsertOnlyReadModel implementation, this is not part of any other readstore implemented.
Can you look at it if you want to include it, I am not sure what the use case should be, you can actually do the same with normal readmodel I am guessing.

wgtmpeters:
@rasmus can you look at my previous comments and check the pull request. I am hoping to have this be pulled in soon

janfeyen:
> @rasmus I have all integrations tests running now, but there is some extra code in the repo that is not tested.
> It is about the IInsertOnlyReadModel implementation, this is not part of any other readstore implemented.
> Can you look at it if you want to include it, I am not sure what the use case should be, you can actually do the same with normal readmodel I am guessing.

I've implemented the IInsertOnlyReadModel back in the day for a specific use case where I did not care about a  read model id as I would never have to be able to get to the read model id again to update the model. I've used it in conjunction with a locator which just generates a new guid:
```

public class NewGuidReadModelLocator : INewGuidReadModelLocator
    {
        public IEnumerable<string> GetReadModelIds(IDomainEvent domainEvent)
        {
            yield return Guid.NewGuid().ToString();
        }
    }
```

The read model itself subscribes to all financial events and because the locator always gives a new guid, we will always end up in a new read model and you will never be able to update the model (hence the IInsertOnlyReadModel).

The use case we had was a financial ledger. The readmodel would hold one and only one transaction for an aggregate and can never be updated once created. It would have fields like the aggregate-id, transaction type, debit / credit, amount etc ....
We then have a query which would return an overview for an aggregate like a bank statement you get from your bank with opening balance and start date, closing balance on end date and all transactions between start date and end date.

I was still fairly new with the concepts of event flow at that time so if there's a better way to do it, please let me know.

rasmus:
I not sure if I'm the right reviewer for this as I have absolutely no experience in working with MongoDB. For me the important part is that the implementation moves in the right direction and that this version is better than that before. You never get it right in the first go, especially with open-source due to the limited amount of time/effort that one can put into the code.

When @wgtmpeters feels like this is ready, I'll merge it in.

wgtmpeters:
@jamfeyen I believe you should be able to do the same with normal readmodels.

The same locator, maybe adding validation in the readmodel that you make sure IsNew is true. Which you should be able to do with the contrext that is injected.

I really want to remove the implementation because it is only implemented in this specific readstore. 

Do you mind if I remove it for now, and if you in the future use this release maybe try to accomplish the same with the normal mongodb readmodel?

wgtmpeters:
@rasmus OK I will finish it soon, we want to use it really soon as well so I will test it in a real project rather quick after release

janfeyen:
@wgtmpeters  you can remove, I'll refactor when upgrading to the latest package

wgtmpeters:
@rasmus I finished the pullrequest.
Created a Initializer class for the setup needed after the resolver has been created.
I think we need to describe these setup examples in the documentation, the same for other implementation that also do some setup that is hidden in the integration test setup. And not really mentioned in the documentation.

When the build is green I am Ok with merging this.

rasmus:
@wgtmpeters good work and thanks for the effort! the documentation is in need of of lot more. the problem, for me at least, is that it takes a lot of time, especially if you want it to be of any quality worth reading.

The documentation is in another repo and it should be relatively easy to extend.

wgtmpeters:
@rasmus will you release soon, we want to take these changes in. At the moment we are working with a custom build version of the MongoDB package.

rasmus:
@wgtmpeters sorry, forgot to get back after I merged to develop. Its released now https://www.nuget.org/packages/EventFlow.MongoDB/

Issue #564, opened 2018-12-02
Bug: LoadAllCommittedEvents fails when there are gaps in GlobalSequenceNumber


frankebersoll:
### Affects
SQL Server since 2012 (happens with Entity Framework, probably other stores, too)

### What I'm trying to do
Apply events to a new Read Model using the `ReadModelPopulator`.

### What I'd expect to happen
It applies all events affecting the Read Model.

### What actually happens
It only applies some events and then stops.

### Cause
SQL Server since 2012 [doesn't guarantee sequential IDs](https://stackoverflow.com/questions/14146148/identity-increment-is-jumping-in-sql-server-database). This can lead to empty pages while enumerating all events, which [stops the population process](https://github.com/eventflow/EventFlow/blob/c852b66f4ea8f2e8efb7f5f748cf704bcb53c218/Source/EventFlow/ReadStores/ReadModelPopulator.cs#L131).

wgtmpeters:
To me having gaps in the sequences does not make sense, but this might also happen in other implementations? If we think it might happen in other specific implementations as well it might make more sense to change the ReadModelPopulator to simply be able to handle this.
In that case we should create a integration test for this bug.

rasmus:
Fixed in #566

PR #566, opened 2018-12-06
Fix LoadAllEventsAsync with gaps (#564)

frankebersoll:
This fixes issues with gaps in Global Sequence Numbers when replaying events. The problem occurs due to database internals like sequence caching, or even by using `IEventPersistence.DeleteEventsAsync`. I used `DeleteEventsAsync` to replicate the problem in our unit tests, but made it optional because I couldn't test all supported DBs.

I found another issue with `EventStoreEventPersistence` tests because the store cannot be reset between test runs. That leads to some tests getting red after repeated runs. I haven't found a way to reliably reset the store.


frankebersoll:
🚢 Ready for shipment.

frankebersoll:
Bonus commits: Fixed all stores.

rasmus:
Great work

PR #579, opened 2019-01-29
add dbo schema

### LINKED ISSUE #578: EventFlow.MsSql - Schema Inconsistent Between Scripts
The script [`0001 - Create table EventFlow.sql`](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow.MsSql/EventStores/Scripts/0001%20-%20Create%20table%20EventFlow.sql) uses the schema `dbo` whereas[ `0002 - Create eventdatamodel_list_type.sql`](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow.MsSql/EventStores/Scripts/0002%20-%20Create%20eventdatamodel_list_type.sql) does not specify any schema. 

I think `0002 - Create eventdatamodel_list_type.sql` should be updated to use `dbo` schema as well to keep it consistent between scripts.

AnthonyCarl:
Keep schema consistent between scripts to resolve #578.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 ### New in 0.69 (not released yet)
 
-* _Nothing yet_
+* Fix: Added the schema `dbo` to the `eventdatamodel_list_type` in script `0002 - Create eventdatamodel_list_type.sql` for `EventFlow.MsSql`.
 
 ### New in 0.68.3728 (released 2018-12-03)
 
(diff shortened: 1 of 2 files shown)


rasmus:
@AnthonyCarl thanks for the PR, could you update the release notes as well?

AnthonyCarl:
I also noticed the [`MsSqlDatabaseMigrator`](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow.MsSql/MsSqlDatabaseMigrator.cs#L41) class does not specifically set `dbo` schema as well. I discovered this because I am running event flow as a user who does not have `dbo` as the default schema and the `SchemaVersions` table gets created in a different schema.

I believe a fix could be as simple as:
`return supportedDatabases.SqlDatabase(connectionString, "dbo");`

I am happy to make a change for that as well if you think it should also be `dbo`. I can do that as part of this issue or start another one.

rasmus:
@AnthonyCarl That would be great. In the future it should maybe be possible to configure, but for now doing the same everywhere is a first step

PR #583, opened 2019-02-07
Fix: Use GetRequiredService in EventFlow.DependencyInjection

### LINKED ISSUE #574: EventFlow.DependencyInjection: Resolve should throw when service is not found
The default container implementation throws an exception when Resolve is called for an unregistered service, which is what you would expect. The ServiceProviderResolver doesn't throw and returns `null` instead.

frankebersoll:
In `ServiceProviderResolver` we used `ServiceProvider.GetService<T>` which returns `null` for unregistered services. In order to align the behavior with other resolver implementations (and user's expectations), this has been changed to use `GetRequiredService<T>` which throws an exception.

This fixes #574.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,8 @@
 ### New in 0.69 (not released yet)
 
 * Fix: Added the schema `dbo` to the `eventdatamodel_list_type` in script `0002 - Create eventdatamodel_list_type.sql` for `EventFlow.MsSql`.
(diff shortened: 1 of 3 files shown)


frankebersoll:
@rasmus This could be merged, but somehow Codacy is questioning the quality of new unit test methods, it wants them to be static. Could be a false positive? I looked into Codacy a bit and couldn't find a way to disable this check for unit test methods. This problem also affects #533

rasmus:
@frankebersoll I removed the Codacy integration.

PR #584, opened 2019-02-07
Added options.RunOnStartup<TBootstrap>()

frankebersoll:
Easily register bootstraps using this extension methods:

```csharp
EventFlowOptions.New.RunOnStartup<DatabaseInitializer>()
EventFlowOptions.New.RunOnStartup(() => Console.WriteLine("started!"))
```

### DIFF
--- a/Source/EventFlow/Extensions/EventFlowOptionsBootstrapExtensions.cs
+++ b/Source/EventFlow/Extensions/EventFlowOptionsBootstrapExtensions.cs
@@ -1,0 +1,68 @@
+﻿// The MIT License (MIT)
+// 
+// Copyright (c) 2015-2018 Rasmus Mikkelsen
+// Copyright (c) 2015-2018 eBay Software Foundation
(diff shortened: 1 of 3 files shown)


Source/EventFlow/Extensions/EventFlowOptionsBootstrapExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+            private readonly Func<Task> _action;
+
+            public ActionBootstrap(Func<Task> action)
+            {
+                _action = action;
+            }
+
+            public async Task BootAsync(CancellationToken cancellationToken)
+            {
+                await _action().ConfigureAwait(false);
```
rasmus: Pass in cancellation token and let developers decide if they want to act on it.

frankebersoll: Fixed

Issue #577, opened 2019-01-28
Async ReadModels (IAmReadModelFor)


JackHerring:
Is there any support for `async` ReadModels? The `IAmReadModelFor<TAggregate, in TIdentity, in TEvent>` interface has only a `void` returning method -> I am loathe to implement as `async void` (not least because fire-and-forget with DB access will inevitably cause problems, but also error-handling and order of execution). I need to do some `async` work before applying my ReadModel changes, though.

Is there a way to do this or is this a case of roll-your-own?

JackHerring:
It looks like `async void` isn't even an option here as the caller of `Apply` will expect the Model to be in a Db-writable state immediately after the call returns which will be at the point of the first `await`. Only way to do this at the moment appears to be blocking on the Task completing with `Wait()`/`GetAwaiter().GetResult()`

rasmus:
@JackHerring the original idea as not to have any complex operations in the read model updates. However, as you have pointed out, this is not always the case and indeed `async void` is a no go as the read model update will continue without the async operation being awaited.

I'll add it as a feature request as see if its possible to do anything about it. It might be that an additional interface as I would really dislike having to introduce such a breaking change to the API.

JackHerring:
@frankebersoll thank you for adding this feature, very much appreciated.

rasmus:
Closing this as its implemented in #585

PR #585, opened 2019-02-08
Async read model updates (IAmAsyncReadModelFor)

frankebersoll:
This adds support for async read model updates #577:

```csharp
public class AsyncPingReadModel : IReadModel,
             IAmAsyncReadModelFor<ThingyAggregate, ThingyId, ThingyPingEvent>
{
    public async Task ApplyAsync(IReadModelContext context, IDomainEvent<ThingyAggregate, ThingyId, ThingyPingEvent> domainEvent)
    {
        await Task.Delay(50);
        PingEventsReceived = true;
    }
...
```

You can mix and match asynchronous and synchronous updates, as long as you don't subscribe to the same event in both ways.

### DIFF
--- a/Source/EventFlow/ReadStores/IAmAsyncReadModelFor.cs
+++ b/Source/EventFlow/ReadStores/IAmAsyncReadModelFor.cs
@@ -1,0 +1,37 @@
+﻿// The MIT License (MIT)
+// 
+// Copyright (c) 2015-2019 Rasmus Mikkelsen
+// Copyright (c) 2015-2019 eBay Software Foundation
+// https://github.com/eventflow/EventFlow
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+// the Software, and to permit persons to whom the Software is furnished to do so,
+// subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


Source/EventFlow/ReadStores/ReadModelDomainEventApplier.cs
```diff
@@ -50,29 +54,73 @@ public class ReadModelDomainEventApplier : IReadModelDomainEventApplier
...
-                                ? null
-                                : ReflectionHelper.CompileMethodInvocation<Action<IReadModel, IReadModelContext, IDomainEvent>>(readModelType, "Apply", methodSignature);
-                        });
+                        return null;
+                    });
 
                 if (applyMethod != null)
                 {
-                    applyMethod(readModel, readModelContext, domainEvent);
+                    await applyMethod.Apply(readModel, readModelContext, domainEvent);
```
rasmus: Missing `.ConfigureAwait(false)`

Source/EventFlow/ReadStores/ReadStoreManager.cs
```diff
@@ -66,6 +66,23 @@ static ReadStoreManager()
...
+            if (AggregateEventTypes.Count != iAmReadModelForInterfaceTypes.Count)
+            {
+                throw new ArgumentException(
+                    $"Read model type '{StaticReadModelType.PrettyPrint()}' implements ambiguous '{typeof(IAmReadModelFor<,,>).PrettyPrint()}' interfaces");
+            }
+        }
+
+        private static bool IsReadModelFor(Type i)
+        {
+            if (!i.GetTypeInfo().IsGenericType)
```
rasmus: Maybe save the type info and check for `IsInterface` to save the call to `GetGenericTypeDefinition` in a few cases

Source/EventFlow/ReadStores/IAmAsyncReadModelFor.cs
```diff
@@ -0,0 +1,37 @@
...
+using EventFlow.Core;
+
+namespace EventFlow.ReadStores
+{
+    public interface IAmAsyncReadModelFor<TAggregate, in TIdentity, in TEvent>
+        where TAggregate : IAggregateRoot<TIdentity>
+        where TIdentity : IIdentity
+        where TEvent : IAggregateEvent<TAggregate, TIdentity>
+    {
+        Task ApplyAsync(IReadModelContext context, IDomainEvent<TAggregate, TIdentity, TEvent> domainEvent);
```
rasmus: I think we should add a `CancellationToken` argument and merely pass a `CancellationToken.None` for now as we open up for IO operations. I hope to be able to add some fine grade controls of when EventFlow passes cancellation tokens through and when it doesn't. 

In most cases you would never want to cancel read model updates, but you might want to if it take too long an blocks application shutdown in a graceful manner.

rasmus:
I also remove Codacy. We did a PoC at work and it didn't really output anything useful.

Source/EventFlow/ReadStores/ReadStoreManager.cs
frankebersoll: It is only used in this context:

```csharp
    .GetInterfaces()
    .Where(IsReadModelFor)
```

Source/EventFlow/ReadStores/IAmAsyncReadModelFor.cs
frankebersoll: Done.

Source/EventFlow/ReadStores/ReadModelDomainEventApplier.cs
frankebersoll: Fixed

PR #587, opened 2019-02-09
Added cancellation boundaries

### LINKED ISSUE #65: Be able to configure "point of no return" for cancellation
Cancellation tokens are used everywhere, but different applications might have completely different views on what is supposed to be cancelled.
- **Likely to have cancellation**
  - Command execution and storage of events in event store
- **Unlikely to have cancellation**
  - Read model updates. The application might depend on read models, so if the event store is updated but the read models isn't, some application might enter a faulty state.
  - Notification of event subscribers

frankebersoll:
Adds a configuration option to set the "point of no return" when using cancellation tokens. After this point in processing, cancellation tokens  are ignored:

`options.Configure(c => c.CancellationBoundary = CancellationBoundary.BeforeCommittingEvents)`

This fixes #65

I went a little overboard with tests, but couldn't find an easier way: Task completion sources control how far the process advances until the token is cancelled. A list of `IStep` instances controls the process and validation.

### DIFF
--- a/Source/EventFlow/Subscribers/DomainEventPublisher.cs
+++ b/Source/EventFlow/Subscribers/DomainEventPublisher.cs
@@ -1,7 +1,7 @@
 ﻿// The MIT License (MIT)
 // 
-// Copyright (c) 2015-2018 Rasmus Mikkelsen
-// Copyright (c) 2015-2018 eBay Software Foundation
+// Copyright (c) 2015-2019 Rasmus Mikkelsen
+// Copyright (c) 2015-2019 eBay Software Foundation
 // https://github.com/eventflow/EventFlow
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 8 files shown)


frankebersoll:
I don't understand the codecov report. How will this commit decrease coverage by over 6%?

rasmus:
It might be due to a missing build on develop. If there isn't an up-to-date report when the build for the PR runs, then the results are a bit off.

Source/EventFlow/Subscribers/DomainEventPublisher.cs
```diff
@@ -91,11 +98,10 @@ public class DomainEventPublisher : IDomainEventPublisher
 
         private async Task PublishToReadStoresAsync(
             IReadOnlyCollection<IDomainEvent> domainEvents,
-            CancellationToken _)
+            CancellationToken cancellationToken)
         {
-            // ARGH, dilemma, should we pass the cancellation token to read model update or not?
```
rasmus: I really, really like that you could remove this 😄

rasmus:
I agree with the test, it doesn't really read well, but better have a complex test than none. I'll sit down with a cup coffee some day and read it thoroughly. I'll merge and release as soon as the build is done

Issue #599, opened 2019-03-13
Failed saga commands are silently ignored


mbican:
AggregateSaga.PublishAsync() doesn't check execution results.
If there is no exception thrown ISagaErrorHandler is not called.
Therefore I suggest to thrown an exception if saga command fails.
I am also creating pull request with my solution.
The Exception Type is open to discussion to integrate it into the framework.

rasmus:
Old issue that has already been fixed. The `PublishAsync` now throws `SagaPublishException` if any commands failed

rasmus:
Fixed in #603

PR #600, opened 2019-03-13
Throw if saga command fails

mbican:
issue #599

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,8 @@ the [do’s and don’ts](http://docs.geteventflow.net/DosAndDonts.html) and the
    [complete example](#complete-example). It features entities, a read model for 
    an entity, delete on read models, specifications and snapshots.
 
- * **[.Net Core:](https://github.com/johnny-chan/EventFlowDemo)**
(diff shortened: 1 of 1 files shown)


Source/EventFlow/Sagas/AggregateSagas/AggregateSaga.cs
```diff
@@ -68,11 +69,20 @@ public virtual async Task PublishAsync(ICommandBus commandBus, CancellationToken
...
             {
-                await unpublishedCommand(commandBus, cancellationToken).ConfigureAwait(false);
+                var executionResult = await unpublishedCommand(commandBus, cancellationToken).ConfigureAwait(false);
+                if (!executionResult.IsSuccess)
+                    fails.Add(executionResult);
             }
+
+            if (0 < fails.Count)
+                if (1 == fails.Count)
+                    throw DomainError.With(fails[0].ToString());
```
rasmus: I would really like to see the failed execution results added to the exception. That way developers get more information on what went wrong and don't need to rely on `.ToString()` on their executions results.

PR #603, opened 2019-03-13
throw on failed saga command

mbican:
I changed branch, issue #599 , original comment:
@rasmus requested changes on this pull request.

Thanks for the fix! Just a few comments

Add { } for if statements
Write a few tests to make sure it doesn't break at some point in the future
Update release notes
As this is a change in behaivor, we need to supply some method of disabling it. It might be a protected property named ThrowExceptionsOnFailedPublish which a default value of true
In Source/EventFlow/Sagas/AggregateSagas/AggregateSaga.cs:

>              }
+
+            if (0 < fails.Count)
+                if (1 == fails.Count)
+                    throw DomainError.With(fails[0].ToString());
I would really like to see the failed execution results added to the exception. That way developers get more information on what went wrong and don't need to rely on .ToString() on their executions results.

### DIFF
--- a/Source/EventFlow/Exceptions/CommandAggregateException.cs
+++ b/Source/EventFlow/Exceptions/CommandAggregateException.cs
@@ -21,40 +21,45 @@
 // IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 // CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-using EventFlow.Aggregates.ExecutionResults;
 using System;
+using System.Collections.Generic;
 
 namespace EventFlow.Exceptions
 {
-    public class CommandException : Exception
+    public class CommandAggregateException : AggregateException
     {
-        public Type CommandType { get; }
-        public IExecutionResult ExecutionResult { get; }
+        public CommandAggregateException()
+        {
+        }
 
-        public CommandException(Type commandType, string message)
+        public CommandAggregateException(IEnumerable<Exception> innerExceptions)
+            : base(innerExceptions)
+        {
+        }
+
+        public CommandAggregateException(params Exception[] innerExceptions)
+            : base(innerExceptions)
+        {
+        }
+
+        public CommandAggregateException(string message)
             : base(message)
         {
-            CommandType = commandType;
         }
 
-        public CommandException(Type commandType, string message, Exception innerException)
+        public CommandAggregateException(string message, IEnumerable<Exception> innerExceptions)
+            : base(message, innerExceptions)
+        {
+        }
+
+        public CommandAggregateException(string message, Exception innerException)
             : base(message, innerException)
         {
-            CommandType = commandType;
         }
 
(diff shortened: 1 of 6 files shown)


Source/EventFlow/Sagas/AggregateSagas/AggregateSaga.cs
```diff
@@ -68,11 +74,37 @@ public virtual async Task PublishAsync(ICommandBus commandBus, CancellationToken
         {
             var commandsToPublish = _unpublishedCommands.ToList();
             _unpublishedCommands.Clear();
+            var exceptions = new List<CommandException>();
 
             foreach (var unpublishedCommand in commandsToPublish)
             {
-                await unpublishedCommand(commandBus, cancellationToken).ConfigureAwait(false);
+                var executionResult = await unpublishedCommand.Item2(commandBus, cancellationToken).ConfigureAwait(false);
```
mbican: @rasmus I have an idea to also catch exceptions from command execution and wrap them into CommandException so developer can just focus on CommandException in ISagaErrorHandler to know which command from saga failed. What do you think?

rasmus: I like the idea as it gives developers more information on which to act. The important this is to make sure that the exception doesn't have a reference to the command, aggregate or any other instance as this would really mess up IoC scopes

rasmus: You could pass the type of the command though, e.g. simply do a `command.GetType()`

Source/EventFlow/Exceptions/CommandAggregateException.cs
```diff
@@ -0,0 +1,65 @@
...
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+using System.Collections.Generic;
+
+namespace EventFlow.Exceptions
+{
+    public class CommandAggregateException : AggregateException
```
mbican: Not the best name 🤔 since there is already an AggregateRoot meaning something else

Source/EventFlow/Sagas/AggregateSagas/AggregateSaga.cs
```diff
@@ -69,10 +77,49 @@ public virtual async Task PublishAsync(ICommandBus commandBus, CancellationToken
...
+                                e));
+                    }
+                }
+                else
+                {
+                    await commandInvoker(commandBus, cancellationToken).ConfigureAwait(false);
+                }
+            }
+
+            if (ThrowExceptionsOnFailedPublish
```
mbican: delete ThrowExceptionsOnFailedPublish from condition

rasmus:
Will merge when build is green. Thanks for the PR

PR #613, opened 2019-03-17
Upgrade ASP.NET Core dependencies

DavidRouyer:
Upgrade ASP.NET Core version to 2.2

### DIFF
--- a/Source/EventFlow.AspNetCore.Tests/EventFlow.AspNetCore.Tests.csproj
+++ b/Source/EventFlow.AspNetCore.Tests/EventFlow.AspNetCore.Tests.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net461</TargetFramework>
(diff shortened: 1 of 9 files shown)


rasmus:
As this is NuGet package the projects merely reference, updating the references only makes sense in two cases.

- Wan't to use features that are new to 2.2 and not in 2.0
- Or, updating to a new major release, e.g. 2.0 to 3.0

If the PR doesn't fall into any of the above, the only thing this actually does, it to force developers to the new versions. They can still update from 2.0 to 2.2 in their own projects.

This is the reason for many of the NuGet package references in EventFlow being quite dated. Its to let developers decide for themselves when to do an update of the references instead of forcing it upon them when they wan't to get the latest EventFlow features.

DavidRouyer:
Maybe we can follow the [.NET Core release lifecycles](https://dotnet.microsoft.com/platform/support/policy/dotnet-core)?
- .NET Core 2.0 is EOL, so I think it's a good practice to enforce developers to use a supported version
- .NET Core 2.1 is LTS <- we pin this one
- .NET Core 2.2 is the current version

rasmus:
@DavidRouyer I agree with the LTS version, EventFlow shouldn't reference anything above if there's no explicit reason.

rasmus:
I'll close this as EventFlow doesn't currently reference any features in 2.2 and thereby let it by up to the developers to decide which version they want

PR #623, opened 2019-04-01
Fix: Index fragmentation using COMB GUIDs in MSSQL

### LINKED ISSUE #622: Using COMB GUIDs in MSSQL causes index fragmentation
As aggregates are stored in MSSQL using their IDs in `AggregateId`, which is a `[nvarchar](255)`, using a COMB GUID doesn't have any positive effect on the index fragmentation. COMB GUIDs only make sense if using the `uniqueidentifier` column type.

rasmus:
Fixes #622


rasmus:
Might be a better idea to support another event store table that has better column types

PR #638, opened 2019-05-15
Add PDB files to packages

rasmus:


rasmus:
Should use https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg

PR #639, opened 2019-05-15
Publishing symbols

rasmus:
AppVeyor doesn't pick up `.snupkg` files according to https://github.com/appveyor/ci/issues/2753.

Documentation here: https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg

EventFlow correctly builds the packages after #638


rasmus:
Publish docs:  https://www.appveyor.com/docs/deployment/nuget/

rasmus:
snupkg seems empty

rasmus:
Although the source link integration is broken, we should still publish the snupkg packages https://github.com/eventflow/EventFlow/issues/644

PR #650, opened 2019-06-09
Rebuild async ReadModels

JackHerring:
I found a bug with using the `IReadModelRepopulator` with async read models, this fixes that problem I think as the problem seems to be in the event discovery when reflecting over the ReadModel (wasn't checking for async ones).

Also deleted an odd looking `TReadModel` class which looks like a refactoring error..? It was preventing me from having a generic argument with the same name and seemed redundant.


rasmus:
👍 thanks for the bug fix

Will merge as when builds are green.

Added missing MIT license headers in accordance with the CLA (some companies care a lot about correct license on projects).

PR #662, opened 2019-07-05
add an external example based on eventflow

twzhangyang:


twzhangyang:
Hi @rasmus 
It looks one test was failed, but I just modified readme.md, any suggestions?

Yang

rasmus:
@twzhangyang some of the RabbitMQ tests are a bit unstable 😢 i haven't been able to identify the problem, even after spending quite a lot of hours on it

rasmus:
Thanks for the link!

PR #665, opened 2019-07-09
Bugfix: Subscribe To All Domain Events

DaGrisa:
Fixes the registration of subscribers implementing the ISubscribeSynchonousToAll interface.


rasmus:
Thanks!

PR #681, opened 2019-09-11
Fix: Deserializing "null" into nullable SingleValueObjects

### LINKED ISSUE #675: Issue in deserializing nullable SingleValueObjects
Hi!

We have defined a `SingleValueObject<int>` in our domain and an event that contains a nullable version of this object. The resulting JSON of this event is either `{ "Key": 42 }` or `{ "Key": null }`.

Serializing to JSON works as expected, but deserializing is an issue because the deserializer does not recognize that the `SingleValueObject` itself is nullable (i.e. it tries to deserialize the value `null` into an `int`, which fails as `int` is a value type).

Let me start with a question: is adding a nullable `SingleValueObject` to an event a bad practice in the first place? I'd like to think it isn't (we've got valid use cases in our domain), but maybe there's something else we can/should do instead.

Second: could it be possible to change the `SingleValueObjectConverter` logic in such a way that it checks for a `null` value first, and deserializes only if a value of type `T` is found?

Related note: an issue might be that you can't distinguish anymore between single value objects that allow for `T = null` _within_ the object, and single value objects that don't. In our case, the fact that `T is int` (and not `int?`) implicates that it should be either instantiated with an int, or not at all -- a solution might be to add a `NullableSingleValueObject`, but it's a bit tedious.

I'd be happy to help with a PR, but like to know other people's thoughts on this first. Thanks in advance!

rasmus:
Fixes #675

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,8 @@
 ### New in 0.75 (not released yet)
 
-* _Nothing yet_
+* Fix: When deserializing the JSON value `"null"` into a struct value like
+  `int`, the `SingleValueObjectConverter` threw an exception instead of
+  merely returning `null` representing an absent `SingleValueObject<int>` value
 
 ### New in 0.74.3948 (released 2019-07-01)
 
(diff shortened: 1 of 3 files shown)


rasmus:
Its getting late here, I'll release this tomorrow

PR #683, opened 2019-09-16
WIP: MSSQL bulk copy

rasmus:
Cherry picking from #464 and inspiration from eBay.


rasmus:
@frankebersoll what di you think about this? I was hoping to use it for huge amount of event imports as some customers would like to move their event sourcing to EventFlow, but require to move their aggregates first.

PR #687, opened 2019-09-25
WIP: C# 8 (nullable)

rasmus:
To do
- [ ] Just make it compile on the main project
- [ ] Remove as many as `#nullable disable` as possible


rasmus:
Moving to .NET instead

PR #693, opened 2019-10-01
General code optimizations.

brunao05:


rasmus:
There's a lot of `System.ObjectDisposedException` exceptions thrown in the failing tests, it might be the same root cause.

brunao05:
I have made a new pull, removing unnecessary changes. Please review.

Source/EventFlow.Autofac/Registrations/AutofacServiceRegistration.cs
```diff
@@ -193,12 +201,11 @@ public void Decorate<TService>(Func<IResolverContext, TService, TService> factor
         public IRootResolver CreateResolver(bool validateRegistrations)
         {
             var container = _containerBuilder.Build();
-            var autofacRootResolver = new AutofacRootResolver(container);
-
-            if (validateRegistrations)
-            {
-                autofacRootResolver.ValidateRegistrations();
-            }
+            using (var autofacRootResolver = new AutofacRootResolver(container))
```
TrayanZapryanov: Is it possible disposed exceptions to be because of this using?

rasmus:
Still failing tests

Source/EventFlow.Autofac/Registrations/AutofacServiceRegistration.cs
FinalFortune: Assume the container gets disposed with it,  not sure why there's a `return new AutofacRootResolver(container);`. 
 Should remove the `using `and have `return autofacRootResolver`

rasmus:
No CLA signed and many of these are already done, closing.

PR #699, opened 2019-10-14
move rabbitmq publish single message to own virtual method 

jepperc:
to allow log instrumentation, e.g. applicationinsights dependency tracking


rasmus:
Thanks

PR #700, opened 2019-10-16
Remove empty constructor requirement on mongo readmodels

ProH4Ck:
Changes already done with other db providers in #393 applied also to MongoDb.


rasmus:
Thanks

PR #686, opened 2019-09-25
.Net core 3 support

ProH4Ck:
This let it works with a .net core 3 project and solves:

`System.TypeLoadException: 'Could not load type 'Microsoft.AspNetCore.Mvc.MvcJsonOptions' from assembly 'Microsoft.AspNetCore.Mvc.Formatters.Json, Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.'
`

PR #709, opened 2019-12-09
Change to VS2019 image

rasmus:
Help out with #686


rasmus:
Seems like the environment override in the YML file has no effect

```
Using connection string for tests: Database=eventflow_2019-12-09-11-27_8d6562cd3e3549b0b682dad946ed1957;Server=(local)\SQL2014;User Id=sa;Connection Timeout=60;Password=Password12!
Executing SQL: CREATE DATABASE [eventflow_2019-12-09-11-27_8d6562cd3e3549b0b682dad946ed1957]
```

rasmus:
Down to a single failing EventStore test. Will have a look at it tonight

Issue #719, opened 2019-12-13
How to unit test CommandHandlers?


Megadesty:
I have a question regarding the testability of `CommandHandler`s. Let's assume I have the classes
```csharp
public class Car : AggregateRoot<Car, CarId> { ... }

public class StartEngine : Command<Car, CarId, IExecutionResult> { ... }
public class StartEngineHandler : CommandHandler<Car, CarId, IExecutionResult, StartEngine> { ... }
```
and I want to unit test the `StartEngineHandler`. One way could be:
```csharp
using (var resolver = this.InitializeEventFlowAndGetResolver()) {
  var bus = resolver.Resolve<ICommandBus>();
  var actual = await bus.PublishAsync(new StartEngine());
```
At this point I can check `actual`, e.g. with `actual.Should().BeOfType<SuccessExecutionResult>();`. But even if that's true, I want to check if the CommandHandler really called the aggregate. So I could do
```csharp
  var queryProcessor = resolver.Resolve<IQueryProcessor>();
  var car = await queryProcessor.ProcessAsync(...);
  car.Should().BeEquivalentTo(new Car(...));
}  
```
because if the read model gets updated, that means the corresponding event was published, which is done by the aggregate, hence the `CommandHandler` called the right thing.

But, that is more an integration than an unit test. So how can I unit-test the `CommandHandler`? Since everything in EventFlow has an interface, I tried to use that to mock the aggregate in the `CommandHandler`'s unit test like this:

```csharp
public interface ICar : IAggregateRoot<CarId> { ... }
public class Car : AggregateRoot<Car, CarId>, ICar {
    public Car(CarId id) : base(id) {}
    ...
}

public class StartEngine : Command<ICar, CarId, IExecutionResult> { ... }
public class StartEngineHandler : CommandHandler<ICar, CarId, IExecutionResult, StartEngine> { ... }
```
This way I can use `StartEngineHandler.ExecuteCommandAsync(new Mock<ICar>())` in my unit test and check afterwards, if the mock was called.

While the unit test is green, the app doesn't work anymore. The problem is, that this exception is thrown:
https://github.com/eventflow/EventFlow/blob/5694823e0b827a7997f05d2955118789bbe33c5a/Source/EventFlow/Aggregates/AggregateFactory.cs#L67-L70
because the interface (`TAggregate` is `ICar`) has obviously no constructor.

[...]

PR #720, opened 2019-12-13
Add example of how to init aggregate's interface

Megadesty:
See EventFlow issue #719

### DIFF
--- a/Source/EventFlow/Aggregates/AggregateFactory.cs
+++ b/Source/EventFlow/Aggregates/AggregateFactory.cs
@@ -57,7 +57,7 @@ namespace EventFlow.Aggregates
             return Task.FromResult((TAggregate)aggregate);
         }
 
-        private static AggregateConstruction CreateAggregateConstruction<TAggregate, TIdentity>()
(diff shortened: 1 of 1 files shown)


rasmus:
@Megadesty the CLA isn't signed, thus I cannot merge the PR 😢

Megadesty:
@rasmus Thanks for looking into this PR and of course I could sign the CLA, but I don't think you want to merge this PR anyway because it's a bad hack which only exist to demonstrate my ideas in question #719. So we can just close this PR after the question was discussed.

rasmus:
Hi @Megadesty, thanks for getting back.

I haven't looked at the code yet as the CLA wasn't signed. I'm afraid that if there's something cool in a PR without the CLA signed that I might be tempted to implement my own version of it, or just be inspired to do something similar down the line.

I really try to do all the "license and copyright stuff" by the book.

rasmus:
I had a look at the PR, and I guess as any good IoC container would throw the same exception with better information, I'll not merge it. The custom made IoC container within EventFlow is up for deletion in version 1.0 anyway (see #724).

Issue #718, opened 2019-12-13
MissingMethodException with .Net Core 3.1 and EventFlow-EntityFramework-Integration


i-dentify:
Now that the .net core 3 PR is merged and available I tried to upgrade my current project. Unfortunately I have a very weird issue, which I cannot debug any further as I cannot get the EventFlow solution running on my MacBook (I'd debug it there, if it was running).

The exception I get is:

```
System.MissingMethodException: Method not found: 'Microsoft.EntityFrameworkCore.Metadata.Builders.IndexBuilder Microsoft.EntityFrameworkCore.Metadata.Builders.EntityTypeBuilder`1.HasIndex(System.Linq.Expressions.Expression`1<System.Func`2<!0,System.Object>>)'.
   at EventFlow.EntityFramework.Extensions.ModelBuilderExtensions.AddEventFlowSnapshots(ModelBuilder modelBuilder)
```
Same goes for "AddEventFlowEvents"...

The strange thing is: if I skip these both calls and ONLY use my own model builder stuff (for my own entities), then there is no such error - though I also call ```HasIndex``` and such.
And to make it even more strange: putting the code from ```AddEventFlowEvents``` directly into my ModelBuilder-stuff does NOT throw this exception.

About the project: it uses .net core 3.1 and event flow with autofac integration and entity framework integration using postgresql. Unfortunately it is quite impossible to strip out the relevant parts as the project already got quite huge and complex. However - according to the .net core 3.0 api changes I had to do some parts different to the EventFlow examples:

AutoFac: no longer integrated within ```ConfigureServices``` but instead within ```ConfigureContainer``` being a void method.

```
public void ConfigureContainer(ContainerBuilder builder)
{
    // ...
    EventFlowOptions.New
                .Configure(options => options.IsAsynchronousSubscribersEnabled = true)
                .UseAutofacContainerBuilder(builder)
                .ConfigureEntityFramework(EntityFrameworkConfiguration.New)
                .UseHangfireJobScheduler()
                .PublishToRabbitMq(RabbitMqConfiguration.With(new Uri(this.Configuration["RabbitMQ:Url"]),
                    exchange: this.Configuration["RabbitMQ:Exchange"]))
                .UseLibLog(LibLogProviders.Serilog);
                //.CreateContainer();
} 
```
Note the point not calling ```CreateContainer()```, which threw exceptions. Obviously .net core 3 does it on its own... or at least the relevant stuff, which is done within this method.

[...]

PR #723, opened 2019-12-30
WIP: Remove .NET Framework

### LINKED ISSUE #713: RFC: Remove .NET Framework
As .NET Framework is rather hard to maintain as it requires Windows to build, I would like some input on whether or not removing .NET Framework from the produced NuGet packages is a good idea at this point.

👍 good idea
👎  bad idea

If you have any comments, please post them below.

rasmus:
Fixes #713


jepperc:
Hi Rasmus. How about going directly to netcore3.1? 3.0 is already going out of service in 2 months. Maybe just both for compatibility :)
Also, it seems lots of microsoft nuget packages also needs updating, like entityframework core. Currently many issues above netcore 2.2 because of these packages, like #718 .
I am available to do some work.. would you rather I work on this, or fix #718 directly first? (it is a blocker for me)

rasmus:
Hi @jepperc, after doing the PR I got to think a lot about how to move EventFlow forward, especially in regards to a 1.0 release.

The current version of EventFlow as several things that I would differently if I were to write it from scratch today, but fixing these issues would introduce a lot of breaking changes. I know that technically EventFlow hasn't yet passed the 1.0 milestone and in respect to SemVer that means your allowed to do basically anything... this doesn't help the many, many projects that are already using EventFlow.

The middle ground is to treat the 0.x releases as a proper release and make the breaking change moving into 1.x.

Some of the topics of the things I would do differently are

- Not having any .NET Framework version, use .NET Standard 2.0 as the lowest API level. Maybe even on build for LTS versions of .NET in the future
- Use Microsoft.Extensions.Logging throughout EventFlow instead of a custom logger and use proper structured logging
- Use Microsoft.Extensions.DependencyInjection throughout EventFlow instead of having a custom implementation
- Remove all the Owin and old non ASP.NET Core packages

These are some of the things I have been thinking about and then on how to move forward in a way that supports that growing community without making the upgrade path too painful. While still making changes that reflect how .NET 5+ applications would be built.

Currently I thinking about writing all of this down in a road map plan and putting it wither on the docs site, this repo or creating a new repo and then get feedback from the community.

rasmus:
Maybe try to gather all this into the existing milestone...

rasmus:
I created https://github.com/eventflow/EventFlow/issues/724 to outline how to move forward, please comment.

rasmus:
Will be handled by #724

Issue #609, opened 2019-03-16
DomainEventPublisher shouldn't stop at exception


mbican:
Since IDomainEventPublisher.PublishAsync() is called after events are already successfully persisted into eventStore all event subscribers should be notified about event. Currently if the first subscriber throws an exception the other subscribers don't receive the event. Better solution would be:
1) catch exceptions from each subscriber
2) collect those exception in an array
3) throw AggregateException() after all subscribers are notified

rasmus:
Think this will be a good candidate for 1.0 as it might rely on changing the code significantly.

Considerations
- I a subscriber throws an exception on the first event, should it still receive the next 
- How should this interact with the `IDispatchToSubscriberResilienceStrategy`

Issue #439, opened 2018-03-06
Crash resilience


rasmus:
If the application using EventFlow shuts down unexpectedly, e.g. due to a power failure, EventFlow will lose consistency between what is committed to the event store and what is applied to read models and subscribers as any in-memory updates will be lost.

While any commands published (passed to `ICommandBus.PublishAsync`) but hasn’t yet resulted in any committed events to the event store are also lost, they are considered out-of-scope for this issue, as EventFlow doesn’t lose consistency.

#### Possible solutions
- Transaction log: Any events committed to the event store could be written to a log and removed when read models and subscribers have been successfully updated
- _others... ?_

#### Mitigation
Steps can be made to minimize the impact that an unexpected shutdown has

- Always drain servers for activity, e.g. web requests, before doing a application shutdown or upgrade
- Use [asynchronous subscribers](http://docs.geteventflow.net/Subscribers.html#asynchronous-subscribers), i.e., let Hangfire (or what ever scheduling you have configured for EventFlow) handle subscriber updates

wgtmpeters:
My team is also looking into situation's like this and 
I believe it would be pretty hard to ensure for every possible crash situation.
Would using asynchronous subscribers not loose the order in which the subscribers are called?
Right now we are sure that the ReadModel Subscribers have been called before any others.


Transaction log you mean that each subscriber has to ACK that it has processed the event?
And the DomainEventPublisher somehow retries after recovery for events that have not been processed by all registered Subscribers?

Has this never been an issue in production @rasmus ?

rasmus:
@wgtmpeters We haven't detected any abnormalities in our production setup that could be traced back to this. However, our production setup is rather stable and not prone to unexpected host crashes or application termination and we carefully drain all servers of web requests before upgrading.

I do think the issue is important, as it will allow me to sleep better at night 😄 (and it makes sense)

For the transaction log, I thought merely to write one entry per committed batch of events and when the `DomainEventPublisher` finishes successfully, then remove the entry. In case of a crash, the events could be picked up and retied. Events would be pushed to read models and subscribers more than once in case of a crash, which would need to be handled.

Even-though our servers in production is stable, our [network infrastructure isn't](https://aphyr.com/posts/288-the-network-is-reliable) and in addition we use RabbitMQ quite a lot. This means that our applications are (usually) written to handle a "at least once event", which is what I thought could be done for read models and subscribers.

wgtmpeters:
[...]

PR #728, opened 2020-01-13
[PoC] reliable event delivery

mbican:
First iteration of Reliable Events for Crash Resilience #439 
Early feedback appreciated

specification originally from a comment https://github.com/eventflow/EventFlow/issues/439#issuecomment-572772200

I have been thinking. We are dealing with the Two Generals Problem ( https://en.wikipedia.org/wiki/Two_Generals'_Problem https://www.youtube.com/watch?v=IP-rGJKSZ3s ) which has been proven unsolvable. The way around it is outbox pattern https://medium.com/engineering-varo/event-driven-architecture-and-the-outbox-pattern-569e6fba7216 . The price that we need to pay for it is to attemp to redeliver events ad infinitum. 

The first step is to extend `IEventPersistance` interface to be able to:
1. list undelivered events
2. mark event as delivered
see suggested extension of interface: https://github.com/mbican/EventFlow/commit/97204b9f30f62a05cccd4d1c4bc03265f316150d

All `IEventPersistance` implementation needs to implement it natively because it needs to be atomic together with event persistance.

It is not only about read models. It is about all receivers of event from `DomainEventPublisher` (read stores, subscribers, sagas). also see #609 for `DomainEventPublisher` modifications 

New event delivery implementation in a somewhat backwards compatible way:
```
# unchanged, commit events
IEventPersistance.CommitEventsAsync(id, ...)

# load undelivered events for aggregate because also some previous events can be still undelivered
var events = IReliableEventPersistance.LoadUndeliveredEventsAsync(id, ...)

# unchanged, publish events
IDomainEventPublisher.PublishAsync(events)

# mark events as delivered (published)
IReliableEventPersistance.MarkEventsDeliveredAsync(id, events)
```
NOTE: We can stop after `IEventPersistance.CommitEventsAsync()` and let events to be delivered asynchronously by the background worker

and/or delivery of events in a background worker in infinte loop:
```
while(true){
    var events = IReliableEventPersistance.LoadAllUndeliveredEvents()
    IDomainEventPublisher.PublishAsync(events)
    IReliableEventPersistance.MarkEventsDeliveredAsync(events)
}
```

So I think the most work will be with all the various `EventPersistance` implementations the change in the EventFlow core should be relatively simple. I can try to provide a prototype of the core change

And yes, it can happen that event will be delivered multiple times (unsolvable Two Generals' Problem). Receiver just has to deal with that e.g. by remembering ids of processed events.
I would even go as far as recommending to call `IDomainEventPublisher.PublishAsync(events)` always twice in Debug build so all code is well tested for duplicate event delivery

### DIFF
--- a/Source/EventFlow.EntityFramework/Extensions/ModelBuilderExtensions.cs
+++ b/Source/EventFlow.EntityFramework/Extensions/ModelBuilderExtensions.cs
@@ -34,6 +34,7 @@ namespace EventFlow.EntityFramework.Extensions
             var eventEntity = modelBuilder.Entity<EventEntity>();
             eventEntity.HasKey(e => e.GlobalSequenceNumber);
             eventEntity.HasIndex(e => new {e.AggregateId, e.AggregateSequenceNumber}).IsUnique();
+            eventEntity.HasIndex(e => e.Confirmed);
             return modelBuilder;
         }
 
--- a/Source/EventFlow/EventStores/IReliableEventPersistence.cs
+++ b/Source/EventFlow/EventStores/IReliableEventPersistence.cs
@@ -8,17 +8,17 @@ namespace EventFlow.EventStores
 {
     public interface IReliableEventPersistence: IEventPersistence
     {
-        Task MarkEventsDeliveredAsync(
+        Task ConfirmEventsAsync(
             IIdentity id,
-            IReadOnlyCollection<IMetadata> eventMetadata,
+            IReadOnlyCollection<IMetadata> eventsMetadata,
             CancellationToken cancellationToken);
 
-        Task<AllCommittedEventsPage> LoadAllUndeliveredEvents(
+        Task<AllCommittedEventsPage> LoadAllUnconfirmedEvents(
             GlobalPosition globalPosition,
             int pageSize,
             CancellationToken cancellationToken);
 
-        Task<IReadOnlyCollection<ICommittedDomainEvent>> LoadUndeliveredEventsAsync(
+        Task<IReadOnlyCollection<ICommittedDomainEvent>> LoadUnconfirmedEventsAsync(
             IIdentity id,
             int fromEventSequenceNumber,
             CancellationToken cancellationToken);
--- a/Source/EventFlow/EventStores/IReliableEventPersistence.cs
+++ b/Source/EventFlow/EventStores/IReliableEventPersistence.cs
@@ -10,7 +10,7 @@ namespace EventFlow.EventStores
     {
         Task MarkEventsDeliveredAsync(
             IIdentity id,
-            IReadOnlyCollection<IEventId> eventIds,
+            IReadOnlyCollection<IMetadata> eventMetadata,
             CancellationToken cancellationToken);
 
         Task<AllCommittedEventsPage> LoadAllUndeliveredEvents(
--- a/Source/EventFlow/EventStores/IReliableEventPersistence.cs
+++ b/Source/EventFlow/EventStores/IReliableEventPersistence.cs
@@ -6,7 +6,7 @@ using EventFlow.Core;
 
 namespace EventFlow.EventStores
 {
-    public interface IReliableEventPersistance: IEventPersistence
+    public interface IReliableEventPersistence: IEventPersistence
     {
         Task MarkEventsDeliveredAsync(
             IIdentity id,
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -58,3 +58,6 @@ If you think your name is missing from the list, create a pull-request.
 
 * MongoDB cleanup and move to main repository
 
+### [Przemysław Andruszewski](https://github.com/przemyslawandruszewski)
+
+* SQL Server and PostgreSQL query generation cleanup
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ share it by creating an issue with the link.
 * **Why _not_ to implement "unit of work" in DDD**
   * [Unit Of Work is the new Singleton](https://blog.sapiensworks.com/post/2014/06/04/Unit-Of-Work-is-the-new-Singleton.aspx)
     by Mike Mogosanu
-  * [The Unit of Work and Transactions In Domain-Driven Design](https://blog.sapiensworks.com/post/2015/09/02/DDD-and-UoW/)
+  * [The Unit of Work and Transactions In Domain-Driven Design](https://blog.sapiensworks.com/post/2015/09/02/DDD-and-UoW)
     by Mike Mogosanu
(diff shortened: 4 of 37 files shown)


rasmus:
@mbican could you update the PR with a short introduction with the general idea?

Source/EventFlow/EventStores/IReliableEventPersistence.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using EventFlow.Aggregates;
+using EventFlow.Core;
+
+namespace EventFlow.EventStores
+{
+    public interface IReliableEventPersistence: IEventPersistence
```
rasmus: Might as well expand the `IEventPersistence` interface with the new methods since its the event store itself that states if its able to handle the requests. Then throw `NotImplementedException` in the event stores that doesn't have the capability.

This would require moving it to 1.0 though...

rasmus:
Another idea would be to have a completely separate interface that stores the log of events.

Basic idea.

1) Store event IDs before they are committed
1) Commit events
1) Publish events
1) Delete event IDs from log

If the service crashes before the events are confirmed, then they are replayed (at least once).

If the service crashes before the events are committed, then any event IDs not in the event store are simply discarded. Events that aren't committed, didn't happen.

This would enable one to implement a log completely separate from the event persistence and would be something we could add to the 0.x release as a default in-memory implementation would simply be a "null implementation".

Thoughts?

mbican:
specification originally from a comment https://github.com/eventflow/EventFlow/issues/439#issuecomment-572772200

I have been thinking. We are dealing with the Two Generals Problem ( https://en.wikipedia.org/wiki/Two_Generals'_Problem https://www.youtube.com/watch?v=IP-rGJKSZ3s ) which has been proven unsolvable. The way around it is outbox pattern https://medium.com/engineering-varo/event-driven-architecture-and-the-outbox-pattern-569e6fba7216 . The price that we need to pay for it is to attemp to redeliver events ad infinitum. 

The first step is to extend `IEventPersistance` interface to be able to:
1. list undelivered events
2. mark event as delivered
see suggested extension of interface: https://github.com/mbican/EventFlow/commit/97204b9f30f62a05cccd4d1c4bc03265f316150d

All `IEventPersistance` implementation needs to implement it natively because it needs to be atomic together with event persistance.

It is not only about read models. It is about all receivers of event from `DomainEventPublisher` (read stores, subscribers, sagas). also see #609 for `DomainEventPublisher` modifications 

New event delivery implementation in a somewhat backwards compatible way:
```
# unchanged, commit events
IEventPersistance.CommitEventsAsync(id, ...)

# load undelivered events for aggregate because also some previous events can be still undelivered
var events = IReliableEventPersistance.LoadUndeliveredEventsAsync(id, ...)

# unchanged, publish events
IDomainEventPublisher.PublishAsync(events)

# mark events as delivered (published)
IReliableEventPersistance.MarkEventsDeliveredAsync(id, events)
```
NOTE: We can stop after `IEventPersistance.CommitEventsAsync()` and let events to be delivered asynchronously by the background worker

and/or delivery of events in a background worker in infinte loop:
```
while(true){
    var events = IReliableEventPersistance.LoadAllUndeliveredEvents()
    IDomainEventPublisher.PublishAsync(events)
    IReliableEventPersistance.MarkEventsDeliveredAsync(events)
}
```

So I think the most work will be with all the various `EventPersistance` implementations the change in the EventFlow core should be relatively simple. I can try to provide a prototype of the core change

And yes, it can happen that event will be delivered multiple times (unsolvable Two Generals' Problem). Receiver just has to deal with that e.g. by remembering ids of processed events.
I would even go as far as recommending to call `IDomainEventPublisher.PublishAsync(events)` always twice in Debug build so all code is well tested for duplicate event delivery

Source/EventFlow/EventStores/IReliableEventPersistence.cs
mbican: agreed,
I split it so it can be compiled with current stores

mbican:
yep log of events, I had a similar idea described here https://github.com/eventflow/EventFlow/issues/439#issuecomment-573874123 it even looks like we can combine both of our ideas.

I see a few disadvantages with log:
1) to be able to successfully recover from the most modes of failure (including e.g. restoring from database backup) it is best to store event-log / outbox in the same database ideally in the same transaction. Because of that I consider extending IEventPersistence interface as superior from the robustness point of view. It will hint developers to implement the event confirmation in the same database.

2)

> If the service crashes before the events are committed, then any event IDs not in the event store are simply discarded. Events that aren't committed, didn't happen.

Problem with that is in case of recovery from a crash. We have deterministic event ids, and the event with the same id will be created again and you will get a conflict between discarding the log of an event that haven't been committed and committing the second event with the same id in the future. Which can be solved by optimistic concurrency and keeping the state of the log that the particular event-id is broken instead of just discarding it.

3) performance - extending IEventPersistence leads to two write operation - 1) commit event 2) confirm event. Whereas log leads to three write operations 1) log event 2) commit event 3) confirm/delete log event

Source/EventFlow.EntityFramework/Extensions/ModelBuilderExtensions.cs
```diff
@@ -34,6 +34,7 @@ public static ModelBuilder AddEventFlowEvents(this ModelBuilder modelBuilder)
             var eventEntity = modelBuilder.Entity<EventEntity>();
             eventEntity.HasKey(e => e.GlobalSequenceNumber);
             eventEntity.HasIndex(e => new {e.AggregateId, e.AggregateSequenceNumber}).IsUnique();
+            eventEntity.HasIndex(e => e.Confirmed);
```
mbican: we also need second index {e.AggregateId, e.Confirmed} for LoadUnconfirmedEventsAsync(id) method

Source/EventFlow/EventStores/IReliableEventPersistence.cs
```diff
@@ -0,0 +1,26 @@
...
+            CancellationToken cancellationToken);
+
+        Task<AllCommittedEventsPage> LoadAllUnconfirmedEvents(
+            GlobalPosition globalPosition,
+            int pageSize,
+            CancellationToken cancellationToken);
+
+        Task<IReadOnlyCollection<ICommittedDomainEvent>> LoadUnconfirmedEventsAsync(
+            IIdentity id,
+            int fromEventSequenceNumber,
```
mbican: remove fromEventSequenceNumber argument, it doesn't make sense in this method

rasmus:
Closing this as it is for the old 1.0 branch

PR #732, opened 2020-01-31
Added quoting to generator of the SQL.

przemyslawandruszewski:
That is the first step I would like to propose for the EventFlow, in the part of generation the SQL. What is does is quoting the column names while generating selects, updates or deletes. Those problems may be related mostly to SQL Server, to the problems while for instance you named your column:"UniqueIdentifier" which is data type in sql or you just have a column in your table that has the same name as system(SQL) tables or their columns.
The second step I would like to propose(if you approve the first one of course) is to propose a bit different way of generating those SQL with library like maybe Dapper but that is after I will make sure it will work for us.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -58,3 +58,6 @@ If you think your name is missing from the list, create a pull-request.
 
 * MongoDB cleanup and move to main repository
 
+### [Przemysław Andruszewski](https://github.com/przemyslawandruszewski)
+
+* SQL Server and PostgreSQL query generation cleanup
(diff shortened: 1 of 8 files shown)


przemyslawandruszewski:
Thanks for PR @rasmus. I've added a word about this change to readme, but probably I will try to write more about this change to the eventflow docs with different PR. As you suggested - added myself to constributors as I would like to help more with development!

rasmus:
Sorry @przemyslawandruszewski I guess I was a bit tired, it meant to update the `RELEASE_NOTES.md` file

przemyslawandruszewski:
Sure :) @rasmus  the release notes were just updated so if you are fine with the changes you can merge it ;)

PR #738, opened 2020-02-04
Added new external example project

OKTAYKIR:


rasmus:
Thanks a lot for this, this is something that several developers have requested

PR #746, opened 2020-03-27
Added ReadModelResult which is an envelope for the result of a read model

Brandon2255p:
Added ReadModelResult which is an envelope for the result of a read model.
When a readmodel is not found then the result is not null but .HasResult can be be used


Brandon2255p:
@rasmus please let me know what you think of this change. I know it would be a breaking change but other than that I would appreciate your thoughts

rasmus:
A very valid point, but as the is a breaking change, we'll have to wait with doing the change until a major increment.

PR #726, opened 2020-01-09
Spike on 1.0

rasmus:
Spike on a version of EventFlow for #724. Nothing is decided yet.

Please note that this isn't very pretty...

## To do

- [x] Delete old projects
- [ ] Make the three core projects work, i.e., EventFlow, EventFlow.TestHelpers, EventFlow.Tests
- [ ] use `Microsoft.Extensions.Logging` everywhere
- Re-add other projects one by one when main project is done
  - [ ] `EventFlow.MsSql`
  - ... the others

PR #754, opened 2020-05-13
PoC: Migrate tests to .NET Core 3.1

rasmus:
Test to see if all tests can be migrated to .NET Core.

@frankebersoll I tried to do everything in one go in #726 and it simply was too much. Too many breaking changes and too much of a big bang while still leaving a lot behind.

Maybe start by slowly introducing .NET Core functionality while still providing backward compatibility.

Personally I hate upgrading NuGet packages across major versions for which the list of breaking changes is a novel in itself.


frankebersoll:
This has been merged together with #782

PR #763, opened 2020-06-17
[Poc] Reimplement asyncsubscribers

leotsarev:


rasmus:
This does however change the order of when they are executed. By moving this code, a subscriber could get invoked before the sync ones are done.

I know its a small possibility, but given a system with a high load, you could end up having your async subscribers getting executed before the sync has completed. Some may rely on this order as it makes it easier to code without having to having too much "eventually consistency".

leotsarev:
@rasmus It this type of breaking change you are not comfortable with?
Another idea could be adding some kind of abstraction over read models, subscribers, sagas... so we could decorate them in common way...

frankebersoll:
@leotsarev This is a breaking change. We cannot treat async subscribers the same as synchronous subscribers, as it invalidates some guarantees about ordering that the framework made before.

In terms of reliable delivery, we may have to consider asynchronous subscriptions separately from the other concepts anyway: While with synchronous subscriptions we have to wait for the handler to complete, with asynchronous subscriptions it is sufficient to wait for successful queuing.

I would really like to see some kind of wiki page that deals with all those expectations and invariants that we already support, and also those that we don't, yet.

@rasmus Would you like to enable the Wiki-Feature? I can imagine that people have a much lower reluctance to contribute their thoughts or knowledge if they just need to slap a little Markdown somewhere. When things have matured, you can then commit them to the documentation repository. What do you think?

leotsarev:
Ok, sadly but true, I understand reluctance to make breaking change

PR #764, opened 2020-06-17
Ms SQL shoud throw OptimisticConcurrencyException on read model insert conflict

leotsarev:
If we got unique constraint violation on inserting read model it probably another thread already inserted read model and we need to start update again.

### DIFF
--- a/Source/EventFlow.MsSql/ReadStores/MssqlReadModelStore.cs
+++ b/Source/EventFlow.MsSql/ReadStores/MssqlReadModelStore.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data.SqlClient;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


leotsarev:
Another option could be to implement retry directly in MsSQLReadStore, but we will lose ability to use RetryPolicies (or will really-really complicate MsSQLReadStore)

rasmus:
If you have a 1-to-1 relationship between your read model and your aggregate, you could try the `AggregateReadStoreManager` which checks if the version is as expected, applies any missing events and the updates the read model.

leotsarev:
Thanks, that's interesting. It's +1 to this implementation (not retrying inside MsSQLReadStore).
However for that to work we need read store to signal that concrete Sql exception is probably OptimisticConcurrencyException, because various generic ReadStoreManager do not know that concrete code of SqlException XXX could be retried specifically in case of INSERT.
So, for me that change is net win.
If you have time, could you please reword your comment to be more actionable (leotsarev, please investigate this, try another approach, also add fix here or there, request review from somebody, add tests...)

rasmus:
@frankebersoll is this still relevant after you merged #768?

leotsarev:
I think it's still relevant. 
**Before change + SingleAggregateReadStoreManager**
If two threads (A & B) will try to create read model in parallel, one of them could fail and lose change.
**Before change + AggregateReadStoreManager**
If two threads (A&B) will try to create read model in parallel, one of them could fail and lose change (say, B).
Sometimes later when another update (C) to this aggregate happen, update B will be added to readmodel in correct order
**After change**
If two threads (A & B) will try to create read model in parallel, one of them could fail and signal OptimisticConcurrencyException. It will be retried.

leotsarev:
@rasmus ?

rasmus:
Still relevant, haven't had the time to give it a go yet though

leotsarev:
Note: I'm not interested anymore in Eventflow because of my job change. I think that PR is still relevant.

PR #765, opened 2020-06-24
[PoC] Add Microsoft.Extensions.Logging without API breaking changes

rasmus:
This PoC is to see if its possible to introduce `Microsoft.Extensions.Logging` without doing any API breaking changes.


Source/EventFlow/EventFlow.csproj
```diff
@@ -31,12 +31,13 @@
...
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
     <PackageReference Include="JetBrains.Annotations" Version="11.0.0">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net452'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
```
leotsarev, line 40: if we are upgrading to nestandard2 as minimum baseline, we could use System.Runtime.Caching from nuget package, if not avail. in target framework.

rasmus:
Doing this in 1.x release already

PR #767, opened 2020-07-03
[ProofOfConcept] crash resilience

MaxShoshin:
## The core problem:
Related #439

We can get in situation when events successfully commited to event store (they committed atomically), but their side effects were not processed (due power failure or error in side effect implementation.).
Probably it's rare situation, but in current implementation there is no way to automatically detect and recover it.

We assume that recovery process for every side effect (`ReadModel`, `Subscriber`, etc) can be differ, so we just detect events which probably side effects were not applied for (or may be only some side effects applied).
Event recovery process has at least once semantic.

**This PR is only to start dicussion, it will be splitted into several PRs if whole concept will be approved.** 

## High level algorithm details:
When side effects is processed successfully we store this fact.
Now we have stream of events in event store and information about which events were published. So we can calculate which events were not published:
We read event from event store and try to find information wether this event was published or not.
If such information found - its ok - this event published successfully, no need to recovery
We have two options if such data not found:
- Side effect is on fly - we should wait and reverify this event bit later
- Probably side effects did not fully apply - we should start recovery process for this event.

## Poof Of Concept implementation details: 
We create general implementation in `EventFlow` project: general classes and interfaces:

We create `PublishVerificator` class - which detects not published events using the `IReliablePublishPersistence` - interface to sotre and read information about published facts. 

`ReliableDomainEventPublisher` - decorator for `IDomainEventPublisher` to store fact of publishing events after the publishing.

Fact storage we implement only based on MS SQL (it should work for all sql like storage, probably similar implementation can be implemented for MongoDb) - we think that it will be enough for check idea and start converstation.

We create two tables `EventFlowPublishLog` (named `PublishLog` in this text below) and `EventFlowPublishVerifyState`.
Last table is used to store state of verification process (`GlobalPosition` to start verify events from event store) - to do not check already verified events from the beginng.

### Publish evens side:
When side effects are applied successfully we add to the `PublishLog` table following info about published events: `AggregateId`, `MinAggregateSequenceNumber`, `MaxAggregateSequenceNumber`. This is rather compact information and using this information we can detect was event published or not.

### Detection of not published events:
Read last verified `GlobalPosition`
Read N events from event store starting from this global position.
Read N oldest records from `PublishLog` *
For each event we try to find `PublishLog` record.
If such record found - its ok - side effect successfully published.
If the record not found we have two options (`IRecoveryDetector` interface is introduced to detect which option it is - it can be tricky in multinode environment ** ):
- Side effect is on fly - we should wait and start verification algorithm from the beginning
- Side effect is not published and should be recovered:
    - pass events to IRecoveryHandlerProcessor who will mark events as successfully published after recovery (via MarkPublishedAsync)
    - go to algorithm beginning to reverify events ***
Compact `PublishLog` table: already verified records can be removed, global position can be adjusted.

#### Some limitation:
- \* If side effect for some event works very slow than we can have N PublishLog records for older events, so we can treat this event as not published. We should add additional check for this case (Not yet implemented in POC).
- \** In POC we use simplest implementation of IRecoveryDetector. We assume that we need to recover event if it doesn't have publish confirmation last 5 minuntes. We have to have better algorithm.
- \*** Algorithm can stuck if IRecoveryHandlerProcessor does not mark recovered events as successfully published. We should have option to stuck and wait for correct recovery algorithm deployed or skip and treat all events after success IReacoveryHandlerProcessor as recovered. It should be user code decision.

#### Other limitations:
- Detection and recovery mechamism in this implementation is not scalable and should work on single node in single instance (its a issue in POC implemntation of verification process, i hope we can solve it). 
- Because of this, verification process can be slower than append new events - we should be able to slow down command execution in this case.
- Distributed lock (to be sure that only one instance of recovery process works in every moment) is not covered somehow, user code should think about it.
- We unable to detect which side effects were successfully published and which were not (for example some events can be successfully aplied for only some of read models). So recovery handler is responsible to handle it.  
- Applied to side effect events can be out of order (correct order is possible only inside single aggregate).
- For single node configuration this algorithm can be improved

#### Other thoughts:
- This implementation doesnot introduce breaking changes (both api and behavior). 
- This solution should work with any event store
- We try to make this algorithm as fast and simple as possible on success way. 
- Fact about success event publish can be stored in defferent way (for example bit column IsPublished on EventFlow table). We have to compare performance.

## Other ideas:
Track position of applied events per side effect.
Pros:
- correct order of applied events
- read model populator works out of the box

Cons:
- breaking change
- less scalability


rasmus:
@MaxShoshin thanks for the PR! this has some really good points and close to what we want.

EventFlow is all about extendibility while providing some sensible defaults. Moving forward with this I think we need to identity which kind of extension points are needed in the core framework to enable implementing crash resilience in different scenarios.

To help identify the API additions needed to EventFlow, a sensible/basic default (and opt-in) mechanic should be implemented.

Again, we need to move forward with this, and from a very short chat @frankebersoll and I had, we'll open a long lived feature branch for which we'll accept PRs. Very similar to what I did a few years back with sagas (https://github.com/eventflow/EventFlow/pull/116). We'll start there, and see how it goes.

As for the content of this PR, it will become a source of reference and may even be a starting point.

MaxShoshin:
@rasmus ,

We would like to contribute in this functionality (We really need it). So we are waiting for new branch.
Then I will create separate PR with change in core API to discuss.

In this PR i've already created basic/default implementation - i think its enough to verify ideas. If not enough - just point me and i'll try to add missing functionality.

In my opinion reliable side effect processing is not possible by default: we need reliable storage for store publishing facts. So i think default implementation should be the same as now - without reliable publishing.

I suggest following high level plan:
- Save information about success publishing
- Detect not published events
- Specific recovery handlers and its default implementation - retry
- Think about multinode recovery process
- Think about smart detector for not published event
- Several recovery strategies: on start only, periodically, lightweight for single node configuration
- Performance tests (what is overhead for write recovery facts, for recovery)

@rasmus , please review this plan/or create your own.

frankebersoll:
@MaxShoshin I like your suggestions, and I'm happy with the general direction of this PR! I'm going to add some code comments.

Source/EventFlow/PublishRecovery/PublishVerificator.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            // Some of not published events can be in flight, so no need recovery them
+            // but we have to check them again on next iteration
+            var eventsForRecovery = GetEventsForRecovery(verifyResult.UnpublishedEvents);
+
+            if (eventsForRecovery.Any())
+            {
+                // Do it inside transaction to recover in single thread
+                // success recovery should put LogItem
+                await _recoveryHandlerProcessor.RecoverAfterUnexpectedShutdownAsync(eventsForRecovery, cancellationToken)
```
frankebersoll, line 72: Isn't this too specific? We don't really know what happened. Maybe just call it "PublishUnconfirmedEvents" or so?

Source/EventFlow/PublishRecovery/IPublishVerificationItem.cs
```diff
@@ -0,0 +1,36 @@
...
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using EventFlow.EventStores;
+
+namespace EventFlow.PublishRecovery
+{
+    public interface IPublishVerificationItem
```
frankebersoll, line 28: IPublishLogItem? IPublishConfirmation?

Source/EventFlow.MsSql/ReliablePublish/MsSqlReliablePublishPersistence.cs
```diff
@@ -0,0 +1,132 @@
...
+
+                await _msSqlConnection.ExecuteAsync(
+                    Label.Named("publishlog-confirm"),
+                    cancellationToken,
+                    "DELETE FROM [dbo].[EventFlowPublishLog] WHERE Id = @Id",
+                    new { logItem.Id});
+            }
+        }
+
+        private sealed class PublishLogItem : IPublishVerificationItem
```
frankebersoll, line 111: This is a DTO/DBO like structure which should contain only properties. Please put the logic somewhere else (extension method or user of this DTO), so IsPublished and IsFinalEvent don't have to be implemented in every persistence provider.

Source/EventFlow/PublishRecovery/IReliablePublishPersistence.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace EventFlow.PublishRecovery
+{
+    public interface IReliablePublishPersistence
+    {
+        Task MarkPublishedAsync(IIdentity aggregateIdentity, IReadOnlyCollection<IDomainEvent> domainEvents);
+
+        Task<VerificationState> GetUnverifiedItemsAsync(int maxCount, CancellationToken cancellationToken);
+
+        Task MarkVerifiedAsync(IReadOnlyCollection<IPublishVerificationItem> verifiedItems, GlobalPosition newVerifiedPosition, CancellationToken cancellationToken);
```
frankebersoll, line 39: Only looking at the interface, I don't understand the difference between `MarkPublishedAsync` and `MarkVerifiedAsync`.

Source/EventFlow/PublishRecovery/RecoveryHandlerProcessor.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            _markProcessor = markProcessor;
+            _readModelRecoveryHandlers = readModelRecoveryHandlers.ToList();
+        }
+
+        public async Task RecoverAfterUnexpectedShutdownAsync(IReadOnlyList<IDomainEvent> eventsForRecovery, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            var recoveryTasks = _readModelRecoveryHandlers.Select(
```
frankebersoll, line 49: Can we safely do that in parallel?

frankebersoll (commented):
This looks promising. However, there are a few points I'd like to address:

- We need to fix the naming. This is something that we need to agree on generally, maybe we can create some kind of glossary. Some examples:
   - Verification / Confirmation / Log
   - Recover / (Re)Apply / (Re)Publish
   - Missing / Failed / Unconfirmed / Unpublished
   - Reliable / Resilient
   - Processor / Handler / HandlerProcessor?

- There are some implementations missing. While there is probably no use for an `InMemoryReliablePublishPersistence`, we probably could need one for faster integration tests. We'd also need an implementation of `IReadModelRecoveryHandler` - or is this supposed to be provided by the user?

- I'm unsure if we can amend this mechanism to store per-subscription information. As I understand it, your solution only stores per-event information, which won't allow us to recover from failures of a single subscriber without affecting all others. You already mentioned that in your initial comment. But will this solution be extendible to provide a more granular support later on?

Thank you for your work!

Source/EventFlow.MsSql/ReliablePublish/MsSqlReliablePublishPersistence.cs
MaxShoshin: Information can be related to storage: for example you can store bit column in EventFlow table. If we agree that we will store some kind of batch/markers i will change this.

MaxShoshin:
So, I'm waiting for #774 and will rebase this PR with code review fixes on top of it.

Source/EventFlow.MsSql/ReliablePublish/MsSqlReliablePublishPersistence.cs
frankebersoll: Yes, you're right - now that I think about it, your solution is more generic. Please ignore the comment.

leotsarev:
@rasmus 
>As I understand it, your solution only stores per-event information, which won't allow us to recover from failures of a single subscriber without affecting all others. You already mentioned that in your initial comment. But will this solution be extendible to provide a more granular support later on?

Yes. You could have implement this strategy on top of recovery handlers. Recovery handler could store information about failure of specific single subscriber into storage of choice and allow some kind of out-of-the-loop (async, scheduled etc) recovery.

leotsarev:
May be we should start splitting into separate PRs during rebase. I.e. start with PR that implements resilience store (IReliablePublishPersistence.cs ) using interfaces from #774

Source/EventFlow/PublishRecovery/IReliablePublishPersistence.cs
MaxShoshin: I've tried to rename methods.

Source/EventFlow/PublishRecovery/IPublishVerificationItem.cs
MaxShoshin: Renamed to IPublishLogItem

Source/EventFlow/PublishRecovery/PublishVerificator.cs
MaxShoshin: Renamed to RecoverUnconfirmedEvents as in recovery strategy we can just skip events (we can decide to do not publish them).

Source/EventFlow/PublishRecovery/PublishVerificator.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        public async Task<PublishVerificationResult> VerifyOnceAsync(CancellationToken cancellationToken)
+        {
+            var state = await _reliablePublishPersistence.GetUnverifiedItemsAsync(PageSize, cancellationToken)
+                .ConfigureAwait(false);
+
+            var logItemLookup = state.Items.ToLookup(x => x.AggregateId);
+
+            var page = await _eventPersistence.LoadAllCommittedEvents(state.LastVerifiedPosition, PageSize, cancellationToken)
```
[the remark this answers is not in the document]
MaxShoshin: I didn't implement high level component yet as it is user specific: IHostedService with infinite loop? Or command line application? Or on start only checker? 

In any case i suppose that this class will change significanlty.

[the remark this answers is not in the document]
MaxShoshin: But yes, when we agree on all related staff i will create high level component with default implementation.

Source/EventFlow/PublishRecovery/ReliableDomainEventPublisher.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using EventFlow.Aggregates;
+using EventFlow.Core;
+using EventFlow.Subscribers;
+
+namespace EventFlow.PublishRecovery
+{
+    public sealed class ReliableDomainEventPublisher : IDomainEventPublisher
```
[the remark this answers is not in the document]
MaxShoshin: Removed. As this implementation rebased on top of development.

Source/EventFlow/PublishRecovery/PublishVerificationResult.cs
```diff
@@ -0,0 +1,32 @@
...
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+namespace EventFlow.PublishRecovery
+{
+    public enum PublishVerificationResult
+    {
+        CompletedNoMoreDataToVerify,
```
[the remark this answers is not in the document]
MaxShoshin: Ok.

Source/EventFlow/PublishRecovery/PublishVerificator.cs
```diff
@@ -0,0 +1,146 @@
...
+        public PublishVerificator(IEventPersistence eventPersistence, IRecoveryHandlerProcessor recoveryHandlerProcessor, IEventJsonSerializer eventSerializer, IRecoveryDetector recoveryDetector, IReliablePublishPersistence reliablePublishPersistence)
+        {
+            _eventPersistence = eventPersistence;
+            _recoveryHandlerProcessor = recoveryHandlerProcessor;
+            _eventSerializer = eventSerializer;
+            _recoveryDetector = recoveryDetector;
+            _reliablePublishPersistence = reliablePublishPersistence;
+        }
+
+        public async Task<PublishVerificationResult> VerifyOnceAsync(CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
MaxShoshin: Renamed to PublishRecoveryProcessor for now, i'll try to look for better name

MaxShoshin:
I'm just rebase current implementation on top of develop branch.
It's still Proof Of Concept. We should play with it and try to understand and select the direction to moove forward.

bmavity-tw:
@MaxShoshin you wouldn't happen to have any updates, would you? Have you found the functionality in this branch to be working for you?

MaxShoshin:
@bmavity-tw , sorry for delay.

We are still in progress. We are only going to use this functionality. 
In any case it is Proof Of Concept and will be redesigned somehow, API will be changed.

bmavity-tw:
Thanks, @MaxShoshin! We'll keep an eye on the progress.

bmavity-tw:
Howdy, again @MaxShoshin. Have you continued to use this implementation or has a redesign begun already? Would you still consider this a Proof Of Concept?

MaxShoshin:
Hi, @bmavity-tw, we still in progress, we still didn't test this in prod.

PR #774, opened 2020-07-08
[PoC] Crash-resilience enabler API

rasmus:
This is a PoC for an API that enables crash resilience as mentioned in #439


Source/EventFlow/Subscribers/ISubscribeTo.cs
```diff
@@ -0,0 +1,29 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+namespace EventFlow.Subscribers
+{
+    public interface ISubscribeTo
```
frankebersoll, line 26: ISubscriber?

Source/EventFlow/Subscribers/IDispatchToSubscriberLog.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using EventFlow.Aggregates;
+
+namespace EventFlow.Subscribers
+{
+    public interface IDispatchToSubscriberLog
+    {
+        Task HandleEventBeginAsync(
```
frankebersoll, line 33: Starting? Started?

Source/EventFlow/Subscribers/IDispatchToSubscriberLog.cs
```diff
@@ -0,0 +1,50 @@
...
+            ISubscribeTo subscriberTo,
+            IDomainEvent domainEvent,
```
MaxShoshin, line 35: I'm affraid that we should store fact information on every event. Can we receive here collection of `IDomainEvent`? And in `HandleEventFailed` method - collection of failed events and in `HandleEventDone` - collection of done events even exception was thrown?

Source/EventFlow/Subscribers/DomainEventPublisher.cs
```diff
@@ -101,7 +105,36 @@ public class DomainEventPublisher : IDomainEventPublisher
...
+                        await _readStoreLog.UpdateBeginAsync(
+                                rsm,
+                                domainEvents,
+                                cancellationToken)
+                            .ConfigureAwait(false);
+                        await rsm.UpdateReadStoresAsync(domainEvents, cancellationToken);
+                    }
+                    catch (Exception e)
+                    {
+                        await _readStoreLog.UpdateFailedAsync(
```
MaxShoshin, line 121: May be it is not log? May be this component can have recovery mechanism? An in this case we should not rethrow exception below?
Moreower, exception here will stop processing other side effects - which may be good in some point, and may be not - it should be user decision.
It will lead to missing call `Begin`... for other logs.. So it will lead to more sophisticated recovery process.

Source/EventFlow/Subscribers/ISubscribeTo.cs
rasmus: `ISubscribe` seems like the right one 😄

Source/EventFlow/Subscribers/IDispatchToSubscriberLog.cs
rasmus: Maybe `BeforeHandleEventAsync` and `AfterHandleEventAsync`?

Source/EventFlow/Aggregates/AggregateStore.cs
```diff
@@ -135,17 +139,42 @@ public class AggregateStore : IAggregateStore
...
+                                aggregate,
+                                commitId,
+                                e,
+                                cancellationToken)
+                            .ConfigureAwait(false);
+                        throw;
+                    }
+                    finally
+                    {
+                        await _eventLog.CommitDoneAsync<TAggregate, TIdentity, TExecutionResult>(
```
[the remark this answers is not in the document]
rasmus: Then rename to `CommitSucceededAsync`?

Source/EventFlow/Aggregates/AggregateStore.cs
```diff
@@ -135,17 +139,42 @@ public class AggregateStore : IAggregateStore
...
+                                sourceId,
+                                cancellationToken)
+                            .ConfigureAwait(false);
+                        return new AggregateUpdateResult<TExecutionResult>(
+                            result,
+                            domainEvents);
+                    }
+                    catch (Exception e)
+                    {
+                        await _eventLog.CommitFailedAsync<TAggregate, TIdentity, TExecutionResult>(
```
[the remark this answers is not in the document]
rasmus: Then move it outside the `try` block?

Source/EventFlow/EventStores/NullAggregateLo.cs
```diff
@@ -0,0 +1,149 @@
...
+            Guid commitId,
+            TExecutionResult executionResult,
+            IReadOnlyCollection<IDomainEvent> domainEvents,
+            Exception exception,
+            CancellationToken cancellationToken)
+            where TAggregate : IAggregateRoot<TIdentity>
+            where TIdentity : IIdentity
+            where TExecutionResult : IExecutionResult
+        {
+            return Task.FromResult(0);
```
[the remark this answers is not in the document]
rasmus: Found the code for the framework version

```
internal static readonly Task<VoidTaskResult> s_cachedCompleted = new Task<VoidTaskResult>(false, default, (TaskCreationOptions)InternalTaskOptions.DoNotDispose, default);
```
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L1446

A lot of internal methods used.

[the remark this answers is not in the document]
rasmus: Created a small internal shim class to wrap it

Source/EventFlow/Subscribers/IDispatchToSubscriberLog.cs
rasmus: Domain events are published to subscribers in order. Thus if an aggregate emits both `EventA` and `EventB`, then `EventA` is sent to all subscribers before `EventB` is sent. We would need to change the order if we wanted to support this.

Maybe something can be done.

rasmus:
@MaxShoshin first thanks for engaging in this discussion! Getting a generic API in the core framework will make it easier to build different strategies on top and hopefully ship a few marked as experimental in the core packages as well.

I'm a bit unsure what you mean by

> I would like to have additional less granular Log interface per DomianEventPublisher:
> I'm afraid that store information per every side effect can be performance costly. So select which interface to implement should decide user: is she wants to have granular information about each side effect processed or is she wants to bit more performance on command execution as they side effects are idempotent.

Could you clarify?

Source/EventFlow/Subscribers/DomainEventPublisher.cs
rasmus: I think that if any recovery/retry is possible, it should be done within the read store as the read store itself has a knowledge of the underlying technology and its capabilities.

The read store is passed as an argument to the log (maybe not the right name as you said) and you could put any custom additional recovery strategies in there.

rasmus:
As @MaxShoshin suggested, maybe the general pattern should be

```csharp
_log.Begin(...)
try
{
  _criticalComponent.Commit(..);
  _log.Succeeded(...)
}
catch
{
  _log.Failed(...)
}
```

frankebersoll:
@rasmus I think what @MaxShoshin meant was that we could have both
- Begin/End for each subscriber: For granular logging of successful publication, while single subscribers may fail and be retried. This might lead to performance problems, as we need to commit for each single subscriber.
- Begin/End per batch of events in the DomainEventPublisher: For logging successfully published events as a whole, if all of your subscribers are idempotent and it doesn't matter to retry all if one fails, saving some performance.

Source/EventFlow/Subscribers/DomainEventPublisher.cs
MaxShoshin: I mean that if side effect throws exception, `_readStoreLog` (and other logs) can handle it in generic way. 

For example:
```
catch (Exception e)
{
    if (! await _readStoreLog.UpdateFailedAsync(....))
    {
        throw;
    }
```

Source/EventFlow/Aggregates/AggregateStore.cs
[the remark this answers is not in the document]
MaxShoshin: Naming things is hard.... :) 
I agree with name your suggested. Its up to you.

Source/EventFlow/Subscribers/IDispatchToSubscriberLog.cs
frankebersoll: In my mind, "Begin" still gets associated with legacy async pattern methods. "Before" sounds better. "After" doesn't express success nor failure.

rasmus:
Hi @frankebersoll and @MaxShoshin I have updated the PR with all the comments

- Reworked when methods are called in relation to the `try-catch` block
- Allowed handlers to disregard the throwing of exceptions if they know how to fix it
- Renamed classes to `[using class name]ResilienceStrategy` as they each has intimated knowledge of the inner workings of the class using it
- Reworked how subscribers are invoked to make sure that all subscribers get invoked even if one of them fails
- ... and some minor cleanup

 There might be some additional details in the dispatch to subscribers that might be useful for a strategy

MaxShoshin:
Hi @rasmus, 

I have some thougths:
- Cancellation: in most cases i don't want to be able to cancel Success/Fail strategy methods. But also I don't think that we can create these methods without `CancellationToken`. May be we should have additional method in `ICancellationConfiguration`?
- It may be usefull if we will have same corellation Id per all strategy actions. Like `commitId` in aggregate store, but for all actions.

CorrelationId usage:
- `IDispatchToReadStoresResilienceStrategy.HandleUpdateFailedAsync` stores info, and return false - to do not rethrow error
- All other side effects (sagas, subscribers) worked
- Called `IAggregateStireResilienceStrategy.EventPublishSucceededAsync' - as no exception is thrown
- Inside this method we can check by correlationId that exception was thrown and raise new exception to indicate that something wrong.

In this case we apply as much side effects as possible. I understand that this algorithm can be dangerous as we violate here rules that subscribers worked after read model updates. But in some cases it can be usefull, so i think user should be able to implement this logic.

Also I think it can be doable with `AsyncLocal` and other stuff, but it  kind of magical approach.
Introduce CorrelationId will lead to a lot of API change. So i suggest to rethink it.

MaxShoshin:
@rasmus ,

In any case, i agree with your changes. I think this api will cover all my crash resilience scenarios.

frankebersoll:
@rasmus @MaxShoshin I'm happy with the additions, and I think we should take some time now to actually use it and gather real feedback instead of trying too hard doing upfront design. As this is relatively new and obviously undocumented, it should be allowed to change in the next few releases, according to the findings of API users. What do you think?

rasmus:
@frankebersoll it put in the release notes that these new API are to be considered experimental

rasmus:
@MaxShoshin I did a small spike on a correlation ID specific for EventFlow in #779

rasmus:
Hi @MaxShoshin 

* For cancellation I think it should be up to each strategy to decide how they want to handle each case if cancellation is requested. As you wrote, it might not make sense to use cancellation is its normal fashion, but it might be good information to the strategy that it should try to persist state or cleanup
* I removed the commit ID but introduced a `BeforeAggregateUpdate` that is invoked at the very beginning. This should make it possible for any strategy to set a correlation/commit on some ambient context before any updates are done. I did a small spike in #779 but I think that should wait a bit and be added in completely new PR

rasmus:
Hi @frankebersoll and @MaxShoshin I think this is a good initial API and as Frank stated, lets see how it works with actual implementations. If you have any final edits or if I didn't get your comments correctly, then ping me. Else give a 👍 and I'll merge it in.

RELEASE_NOTES.md
```diff
@@ -9,6 +9,14 @@
 * Fixed: An issue where `EntityFrameworkEventPersistence` could possibly save aggregate 
   events out of order, which would lead to out-of-order application when streaming events
   ordered by GlobalSequenceNumber
+* New: A new set of hook-in interfaces are provided from this release, which should
+  make it easier to implement crash resilience (#439) in EventFlow. Please note that
+  this new API is experimentation is subject to change as different strategies are
```
frankebersoll, line 14: "is experimentational and subject to change"

PR #775, opened 2020-07-08
file event store will now use relative paths

### LINKED ISSUE #771: allow FilesEventPersistence to use relative paths
Currently `FilesEventPersistence` is using absolute paths to store events. If the user changes `IFilesEventStoreConfiguration.StorePath` and copies the DB to the new locations (for example while deploying to another server or reorganizing data paths during development), events are not deserialized properly, as the event files no longer exist in the old location - all the events are effectively unreadable and lost. Furthermore, there are no warnings raised, which makes debugging the missing events difficult.
This could be easily fixed if the paths were relative to `IFilesEventStoreConfiguration.StorePath` - the `StorePath` is used to point to both event log (`Log.store` file) and events themselves, so they'll always reside in the same location.

I will probably submit a PR for this, but I'd like to hear your comments, especially if you think this is a bad idea.

BTW, I'm aware that `FilesEventPersistence` shouldn't be used in production environment, but it's sometimes more convenient during development - at least until you reorganize your data paths, which can happen during development cycle.

Yaevh:
fixes #771


Source/EventFlow/EventStores/Files/IFilesEventLocator.cs
```diff
@@ -29,5 +29,6 @@ public interface IFilesEventLocator
     {
         string GetEntityPath(IIdentity id);
         string GetEventPath(IIdentity id, int aggregateSequenceNumber);
+        string GetRelativePath(string relativeTo, string path);
```
frankebersoll, line 32: While the calculation of entity and event paths can change, this method always does the same thing. It doesn't belong in this interface. Maybe just put it in FilesEventPersistence, where it is used?

frankebersoll:
I'm going a bit havoc with the projects feature. Sorry about that!

PR #779, opened 2020-07-12
[PoC] Event correlation

rasmus:
PoC of event correlation for #774. `AsyncLocal<>` was introduced in .NET 4.6, thus the need to upgrade.


MaxShoshin:
@rasmus , 

If you want to implement this without upgrade you can use CallContext for old frameworks.
I've see this into:
https://github.com/serilog/serilog/blob/dev/src/Serilog/Context/LogContext.cs

rasmus:
Too old

PR #780, opened 2020-07-12
Obsolete AsyncHelper

rasmus:
Of all the things we can put in a 1.0 release, this is in the very top. The others are the deletion of the custom IoC container and logger.


Source/EventFlow/Extensions/EventStoreExtensions.cs
```diff
@@ -56,14 +58,15 @@ public static class EventStoreExtensions
...
+        [Obsolete("Non-async extensions methods will all be removed in EventFlow 1.0, use async methods instead")]
         public static AllEventsPage LoadAllEvents(
             this IEventStore eventStore,
             GlobalPosition globalPosition,
             int pageSize)
         {
             return eventStore.LoadAllEvents(globalPosition, pageSize, CancellationToken.None);
         }
-
+        [Obsolete("Non-async extensions methods will all be removed in EventFlow 1.0, use async methods instead")]
```
frankebersoll, line 69: \r\n 😉

PR #782, opened 2020-07-15
.NET Core 3.1 Tests

frankebersoll:
Related to #781

## Summary

Our tests have been updated to use the .NET Core Runtime 3.1. This also required some breaking changes in Entity Framework, Hangfire and MongoDB.

### Runtime and TargetFramework

All test project files previously were targeting the .NET Framework (`net452` or `net461`). They have been updated to use the target framework `netcoreapp3.1`, which means that tests will be executed with .NET Core.

There is one exception: The OWIN test project now uses `net472`, as OWIN doesn't exist in ASP.NET Core.

Using the .NET Core runtime implies that tests can be discovered and run either by the Visual Studio Test Explorer, the ReSharper Test Runner, `dotnet test` or any VS Code extensions that use the dotnet test infrastructure. It also means [that the NUnit Console runner won't work any more](https://github.com/nunit/nunit-console/issues/475).

### NUnit Framework and Adapter references

The test projects used to solely reference `NUnit3TestAdapter`, which turned out to be insufficient for reliable discovery by all test runners. New test projects created using the Visual Studio templates also reference `nunit` and `Microsoft.NET.Test.Sdk`, those were added to all our project files.

All test framework references were updated to the latest version, while any obsolete and unused references to other test frameworks or libraries were removed.

## Specific projects

✅ - Non-breaking change outside of test projects

🔺 - Possibly breaking change outside of test projects

### ASP.NET Core

- ✅ Added `netcoreapp3.1` to the supported frameworks (and updated all conditional compilation directives).
- The tests were updated according to [the ASP.NET Core 3.1 migration guidelines](https://docs.microsoft.com/de-de/aspnet/core/migration/22-to-30).
- Reorganized all package references.

### Entity Framework

- 🔺 Tests and library were updated from EF 2.2.6 to 3.1.5. [This might introduce breaking changes to users](https://docs.microsoft.com/de-de/ef/core/what-is-new/ef-core-3.0/breaking-changes), while our library doesn't have any API changes itself.
- This update requires us to denote our test read model id properties with `ValueGeneratedOnAdd`, described [over here](https://docs.microsoft.com/de-de/ef/core/what-is-new/ef-core-3.0/breaking-changes#string-and-byte-array-keys-are-not-client-generated-by-default).

### Hangfire

- 🔺 `IHangfireJobRunner.Execute : void` was changed to `ExecuteAsync : Task`. This might be a breaking change, but hopefully there aren't too many implementers and updating should be trivial.
- The previous test implementation made a request to the Hangfire dashboard in order to figure out if a job was run successfully. Now, we use the `IServerFilter` extension point to get notified about any job status updates. This means we can drop any HTTP references and just use the in-memory `BackgroundJobServer`.

### MongoDB

- 🔺 `MongoDB.Driver` was updated from 2.7.2 to 2.10.4. This introduces some minor API changes, but no incompatibility to older MongoDB versions.

### SQLite

- ✅ `System.Data.SQLite.Core` was updated from 1.0.109.1 to 1.0.113.1

### SQL

- ✅ Now using `System.ComponentModel.Annotations` version 4.7.0 when referencing `netstandard2.0`.

### TestHelpers

- Added [`Appveyor.TestLogger` in order to simplify AppVeyor output](https://github.com/spekt/appveyor.testlogger).
- Fixed a bug where it wasn't possible to run tests on a local MSSQL server instance with Integrated Security.

### CodeStyle

- Added two new projects `EventFlow.CodeStyle` and `EventFlow.CodeStyle.Tests`, containing roslyn analyzers that replace any tests that do statical analysis (e.g. checking for test category attributes or license header text).
- Added an analyzer and code fix that checks for missing `CategoryAttribute` on all test classes. It also validates the argument to be one of the `Categories` constants.
- Added a `Directory.Build.props` [that gets automatically included in all projects](https://docs.microsoft.com/de-de/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets). This conditionally adds the analyzer to all test projects, [using a technique from StackOverflow](https://stackoverflow.com/a/57435187/4317569). We can also use this file instead of the existing `Common.props`, but I didn't want to touch all projects in this PR.
- The analyzers produce errors that show up in Visual Studio and also when building in the command line with `dotnet build`.

## Build process

Testing from the command line now requires using `dotnet test`. The many changes to the build script reflect this:

- `NUnit.ConsoleRunner` was replaced by `DotNetCoreVSTest`.
- There is no support for logging the raw console output to a text file, as there was in the NUnit console runner. We use the [HtmlLogger](https://github.com/microsoft/vstest/blob/master/src/Microsoft.TestPlatform.Extensions.HtmlLogger/HtmlLogger.cs) instead.
- Logging to a file and uploading it to AppVeyor seemed cumbersome, so we just use the [`Appveyor.TestLogger`](https://github.com/spekt/appveyor.testlogger). This has the advantage that test results show up in real time in AppVeyor.
- The loggers (and any NUnit specific switches) cannot be configured properly using command line arguments, so we use a `Test.runsettings` file and pass it as an argument.
- The test runner gets called once for each target runtime: `net472` first, then `netcoreapp3.1`, as the runner cannot process DLLs with different runtimes at the same time. 

There also have been some minor improvements to the build process, including:

- `NoRestore` on build, because we depend on restore.
- OpenCover support was fixed: There was a problem when trying to register the OpenCover DLL. 

## Miscellaneous

- Ignored .ionide and .vscode directories to better support VS Code.


rasmus:
@frankebersoll just got back from my vacation, I'll make this my top priority. On the surface it looks very nice and a lot of healthy improvements.

rasmus:
I'll merge it in to get the benefits of the updated tests and the release notes can come later

Issue #465, opened 2018-05-18
Support Kafka


rasmus:
Until Docker integration testing is ready (#425), maybe setup a free plan using https://www.cloudkarafka.com/plans.html

Use https://github.com/confluentinc/confluent-kafka-dotnet as its the recommended client from Microsoft.

RTodorov:
Hi @rasmus,

Looks like EventFlow is the most mature DDD/CQRS/ES library for C# out there, kudos for that!

I'm looking into adopting it on different projects but I'm gonna be using Kafka as the main event store / service bus. Any idea how hard would be to implement Kafka support on my own? Or is there any active workstream for adding kafka support other than this issue?

Thank you

rasmus:
@RTodorov the persistence layer in EventFlow is defined by the [`IEventPersistence`](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow/EventStores/IEventPersistence.cs) which isn't that complex.

You could get some inspiration for how to implement it by looking at some of the existing implementations, e.g. the [`InMemoryEventPersistence`](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow/EventStores/InMemory/InMemoryEventPersistence.cs)

If you do select to do the implementation, then the event persistence test suite that EventFlow uses to tests its implementations are shipped in the [`EventFlow.TestHelpers`](https://www.nuget.org/packages/EventFlow.TestHelpers/) NuGet package, specifically the class `TestSuiteForEventStore`. Setting it up is rather simple, have a look at the [`MsSqlEventStoreTests`](https://github.com/eventflow/EventFlow/blob/develop/Source/EventFlow.MsSql.Tests/IntegrationTests/EventStores/MsSqlEventStoreTests.cs) to get an idea on how to get started.

If you get a working implementation along with a suite of integration tests, then we could add it to the repo.

Hope that helps.

ahernandez-allegion:
@rasmus I'm curious, would supporting a basic Kafka extension really be that different than the existing RabbitMQ extension?

PR #791, opened 2020-08-26
[PoC] Adding Kafka as Event Publishing Alternative to Rabbit

Herlitzd:
Related #465

Added an integration with Kafka using the `Confluent.Kafka` package.

I heavily followed the patterns established from the RabbitMq integration. At the moment, my testing has been limited to the integration testing that is included with this PR.

I would love to get some feedback, and discuss any changes that would need to be made in order for this integration to move forward.

#### Notes
* Due to the complexity of the Kafka configuration, I am leaking the Producer Configuration object out through the `IEventFlowOptions` builder:
    ```
    .PublishToKafka(new ProducerConfig { })
    ```
* In order to allow consumers to customize their partitioning scheme I am allowing the Kafka publisher to be configured with an optional `Func<IDomainEvent, TopicPartition> topicPartitionFactory`

Edit: Looks like the integration test is timing out, unfortunately a Kafka broker doesn't expose any kind of meaningful health check, at least not without adding a third container for diagnostics. I can try to crank up the timeout in the integration test, but I would rather not burn through your CI time if there are more structural concerns with this change to address first.


ahernandez-allegion:
I'll be following this and may have my team take a look. Thanks for taking a stab at it.

PR #810, opened 2020-12-04
Firsts steps toward 1.0 alpha/beta

### LINKED ISSUE #809: Poll: Drop .NET Framework support?
Having to support .NET Framework makes working with the EventFlow code base quite a hassle, integrating it with new modern services that use .NET (Core) even more so.

This is a poll to get a feeling for how the community thinks in regards to dropping support for NET Framework completely and only supporting .NET Standard 2.0 and onward.

Please use emojis to vote.

👍  best idea ever
👎  please no, its critical for us
😕  not great, but we'll manage
 
Please leave comments if you have something on your mind.

rasmus:
See how the EventFlow might look like without .NET Framework

Fixes #809


rasmus:
Merge to get the updated README

PR #813, opened 2020-12-07
v1: Fix build

rasmus:


rasmus:
Trying GitHub actions in #814 to see if its easier to fix there and if there's a better queue time for builds

rasmus:
Builds and tests correctly on #814 😢 expect some problem with loaded assemblies

rasmus:
Seems like the problem is with the AppVeyor setup, disregarding for now and relying on GitHub actions. Actions also seem to have a considerably lower queue time

Issue #807, opened 2020-11-19
Add support for specific ISagaErrorHandler<ISaga>


ProH4Ck:
For a project I'm working on an implementation of ISagaErrorHandler to allow use of different ISagaErrorHandler for each Saga defined in the project.

Here is a little concept:

```csharp
public interface ISagaErrorHandler<TSaga> : ISagaErrorHandler
  where TSaga : ISaga
{
}
```

```csharp
public class GenericSagaErrorHandler : ISagaErrorHandler
{
    private readonly IResolver _resolver;

    public GenericSagaErrorHandler(IResolver resolver)
    {
        _resolver = resolver;
    }

    public async Task<bool> HandleAsync(ISagaId sagaId, SagaDetails sagaDetails, Exception exception,
        CancellationToken cancellationToken)
    {
        Type specificSagaErrorHandlerType = typeof(ISagaErrorHandler<>).MakeGenericType(sagaDetails.SagaType);
        var specificSagaErrorHandler = (ISagaErrorHandler) _resolver.Resolve(specificSagaErrorHandlerType);
        return await specificSagaErrorHandler.HandleAsync(sagaId, sagaDetails, exception, cancellationToken);
    }
}
```

Code is not tested because still a work in progress!

Do you think this could be integrated in the library? Maybe added to DefaultSagaErrorHandler? 

PS: I used `IResolver` just to keep this aligned with other EventFlow classes but this is an anti-pattern... it should be replaced with a factory (implemented via `Func<>`). This could be another step for 1.0 😉

WardenUnleashed:
I've run into this issue as well but solved it somewhat differently.

I have implemented the following classes to handle Sagas by type

```c#
public class AggregateSagaErrorHandler : ISagaErrorHandler
    {
        private readonly ConcurrentDictionary<Type, ISagaErrorHandler> _errorHandlers;

        public AggregateSagaErrorHandler(IEnumerable<KeyValuePair<Type, ISagaErrorHandler>> errorHandlers = null)
        {
            if (errorHandlers == null) errorHandlers = Enumerable.Empty<KeyValuePair<Type, ISagaErrorHandler>>();
            _errorHandlers = new ConcurrentDictionary<Type, ISagaErrorHandler>(errorHandlers);
        }

        public void RegisterSagaErrorHandler<T>(ISagaErrorHandler<T> errorHandler)
            where T : ISagaId
        {
            _errorHandlers[typeof(T)] = errorHandler;
        }

[...]

PR #817, opened 2020-12-10
Feature/generic saga error handler

ProH4Ck:
Implements #807

### DIFF
--- a/Source/EventFlow/Sagas/SagaErrorHandler.cs
+++ b/Source/EventFlow/Sagas/SagaErrorHandler.cs
@@ -29,9 +29,9 @@ namespace EventFlow.Sagas
 {
     public class SagaErrorHandler : ISagaErrorHandler
     {
-        private readonly Func<Type, ISagaErrorHandler<ISaga>> _sagaErrorHandlerFactory;
+        private readonly Func<Type, ISagaErrorHandler> _sagaErrorHandlerFactory;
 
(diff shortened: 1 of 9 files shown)


Source/EventFlow/Sagas/SagaErrorHandler.cs
```diff
@@ -29,14 +29,29 @@ namespace EventFlow.Sagas
...
+
         public Task<bool> HandleAsync(
             ISagaId sagaId,
             SagaDetails sagaDetails,
             Exception exception,
             CancellationToken cancellationToken)
         {
-            // The default handler cannot handle anything!
+            // Search for a specific SagaErrorHandler<Saga> based on saga type
+            ISagaErrorHandler sagaErrorHandler = _sagaErrorHandlerFactory(sagaDetails.SagaType);
```
rasmus: @ProH4Ck I think this should be moved to the `DispatchToSagas` instead. Doing it here makes it harder to do a "catch all" error handler as well as you will need to copy the code to handle the specific error handlers yourselv.

rasmus:
@ProH4Ck ❤️ thanks for the PR

Source/EventFlow/Sagas/SagaErrorHandler.cs
ProH4Ck: Correct. I'll push changes to the PR in the evening

rasmus:
❤️

PR #819, opened 2020-12-29
Query Processing IoC Refactor

WardenUnleashed:
I have been working on integrating [SignalR with EventFlow](https://gitlab.com/WardenUnleashed/eventflow-signalr) and was able to get it working at a rudimentary level.

The integration required some pre-work to get everything working correctly and that work is some that I felt others could benefit from!

I'm hoping I can get some feedback / confirmation this is change that would be desirable. If so, I'll add the relevant tests to it!

Getting this in for v1 would be ideal since it has breaking changes to queries(namely the addition of a method that works similar to the one on the `Command` class)

Additionally, I noticed that the `ISerializedCommandPublisher` only returns the `ISourceId` while consuming the `IExecutionResult`. It made me wonder if instead it should return an `IExecutionResult` which in turn should have the `ISourceId` from the command appended to it.
This way, you could always link a command to its results! What do you think?


rasmus:
Hi @WardenUnleashed this looks interesting and it would make it easier to integrate queries where there's no generic type

WardenUnleashed:
Added a refactor of the query processor's IoC. Since this would be a breaking query change any, I feel like rolling them together would be good!

rasmus:
Hi @WardenUnleashed 

Have a look at what I did in #825, it implements the method that you were missing on the `IQueryProcessor`, without any need to change existing queries and adding another method on the `IQueryHandler`.

I do believe that EventFlow should try to keep all the nasty stuff bundled within itself and try to keep the interfaces that developers need to implement as simple as possible. Thus adding an extra method `IQueryHandler` merely to make it easier to implement in EventFlow, isn't a good solution.

Last time I some tests on the `ReflectionHelper.CompileMethodInvocation` implementation, it was 400-600 times as efficient as doing a `MethodInfo.InvokeMethod` call, as it generates IL code and compiles it runtime, basically its a very efficient type cast.

Maybe it can be used to remove the `task.GetType().GetTypeInfo().GetProperty("Result").GetValue(task)` part from the PoC I suggested.

If the implementation from my small PR was added, then no breaking changes are needed (as far as I can see) and we could add this feature to the 0.x release.

WardenUnleashed:
Nice! That seems like the better option for now! I'll refactor my branch over this week to get it in line!

WardenUnleashed:
@rasmus Finally was able to get back around to this! Not so many changes required and looks like it could be added to v0! Also added unit tests!

rasmus:
Nice, I'll have a look

WardenUnleashed:
@rasmus Just giving you another ping to see if this can go into develop-v1 soon :D

rasmus:
@WardenUnleashed there seems to be a few tings wrong with the branch, spent a lot of time this evening trying to get it working on AppVeyor. I'll try to get it working asap

Issue #820, opened 2020-12-29
Lack of event segregation corrupts event store


Reyhn3:
Using `IIdentity` when persisting events can lead to overlapping identities between **different** identity types, causing events and aggregates to be corrupted.

For example, if `LocationId` has the value of `abc`, and the `VoyageId` also has the value of `abc`, then the `LocationCreatedEvent` and `VoyageCreatedEvent` will collide. Both events are valid and can be persisted, but when loading the events, an `InvalidCastException` will be thrown for both the location and the voyage.

The documentation does not state that an implementation of `IIdentity` **must** be globally unique (or pre-partitioned). I, personally, do not think that the `IIdentity` should be responsible for ensuring global uniqueness or partitioning. I suggest that responsibility is owned by the `IEventPersistence` who can better determine how to properly persist and retrieve events related to a specific aggregate and persistence strategy. However, the `Identity<T>` class (that is very well documented) currently has this logic, for what I assume are good reasons. As such, I think that particular logic should remain intact, but the `IIdentity`-interface should still be fully honored.

The problem is that the `EventStoreBase` does not currently provide any aggregate type information to the `IEventPersistence`. (Note that the `IReadModelStore<T>` does this, as well as the `ISnapshotPersistence`.) Instead, the `EventStoreBase` implicitly relies on the highly specific logic inside `Identity<T>` to provide a partitioned event record key based on the `IIdentity`-implementing class. The problem remains hidden as long as you use `Identity<T>`.

The challenge is that this problem affects all current event stores, which store data differently. Any fix to actually solve the problem would need to update the store implementations individually **and** migrate all existing event data.

WardenUnleashed:
The interface `IIdentity` to me intrinsically implies a globally unique identity due to the lack of typing/additional information within the interface. It quite literally is just a string value. If I were to try and use these values as keys to a dictionary  I would naturally expect there to be a collision if an identity of `abc` existed and I tried to add another `abc` entry.

Consider the following identity implementation
```c#
public class StringIdentity : IIdentity
    {
        public StringIdentity(string value)
        {
           Value = value
        }

        public string Value {get;}
    }
```
If we take a look at your unit test example, I would argue that your `FirstAggregateId` are `SecondAggregateId` identities are equivalent to this `StringIdentity` class and thus it would be natural for a FirstAggregateId with a value of `abc` to collide with a SecondAggregateId also with a value of `abc`.

[...]

PR #821, opened 2021-01-05
Event segregation by aggregate type

Reyhn3:
This PR addresses the issue that `IIdentity` is not fully honored by `IEventPersistence`, as reported in  #820.

Without this fix, it is not possible to implement your own `IIdentity` because identities of different aggregate **types** risk colliding. The changes in this PR are related to adding aggregate type info to event stores, so that they may store events from different aggregate types smarter.

### Breaking changes
The `EventStoreEventPersistence` for **Event Store** has had the stream names prefixed with the aggregate type. That will make this code unable to find any existing streams.

### Caution advised
The **SQL**-based event persistence libraries already stored the aggregate type (name). The loading of events have been updated to include the aggregate name as a criteria, which *may* affect performance. The table creation scripts have had the index updated to compensate for this, but current tables may have to be manually reindexed.

This also affects **EntityFramework**, depending on which database is used.

As for **MongoDB**, the aggregate name has been added to the index. I am unsure of what impact this may have on existing databases.


rasmus (changes requested):
Thanks for the PR ❤️ 

As you pointed out, this PR contains a lot of breaking changes and we cannot include it in the 0.x release. While 0.x denotes a "beta" release, starting to do major breaking changes here wouldn't be fair for the users of EventFlow.

While I do think that this area is due for a cleanup, doing data breaking changes is something that will never make it into the EventFlow code base as this would make it virtually impossible for anyone with terabytes of data to do live updates across major versions of EventFlow.

If the structure that the underlying store has needs to change in breaking manner, which I do think should be supported, we'll need to implement some method of selecting a scheme when configuring the event persistence, e.g. the MSSQL persistence might have have a scheme for which aggregate and event IDs are GUIDs and another using a different string type or even some custom scheme controlled outside EventFlow.

For this PR, I think the first step is to make sure that the aggregate type is parsed to the persistence layer and do it in the `develop-v1` release.

Issue #678, opened 2019-09-04
Replace JSON.NET with System.Text.Json


DavidRouyer:
.NET Core 3.0 is coming with the new [System.Text.Json API](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-apis/) with much better performance than JSON.NET, what do you think of replacing it with the new API?

rasmus:
You can use it today already, simple provide an implementation that implements `IJsonSerializer` and EventFlow will use that instead.

Right now EventFlow targets .NET Framework 4.5.2 and System.Text.Json requires 4.6.1. I'm not sure that increasing the framework requirement for improved performance is a good idea, especially when you can change the implementation.

Might consider using different JSON implementations for different frameworks, i.e., System.Text.Json for .NET Standard and then JSON.NET for .NET Framework.

DavidRouyer:
> Might consider using different JSON implementations for different frameworks, i.e., System.Text.Json for .NET Standard and then JSON.NET for .NET Framework.

Looks good to me 👍

bartelink:
One thing to bear in mind is that STJ doesnt necessary have 100% compatibility with json.net conventions and doesn't honor json.net Converters that one might be reliant on. The choice of serializer and settings is definitely not something one'd want to happen without an explicit change somewhere - i.e. I'd expect to at least need to change a line of code to opt in.

rasmus:
@bartelink agree

tedvanderveen:
I have submitted PR #760 that allows for more flexible use of serializers, basically not just lock in one string based serialization. 
I'm very much interested in your feedback.

Yaevh:
how are things progressing re System.Text.Json?
I am corrently working on a project that uses System.Text.Json as primary JSON serialization library, but I still have to duplicate large portions of JSON serialization/deserialization logic to be able to use EventFlow, which is kinda pain in the butt.

rasmus:
@Yaevh should implementing `IEventJsonSerializer` be enough?

rasmus:
But it something I would like to do for 1.0

Yaevh:
I've been playing around with System.Text.Json and noticed two differences from Netwonsoft.Json:
1. System.Text.Json treats empty JSON as invalid and throws `JsonException`; Newtonsoft.Json deserializes it to `null`
2. Newtonsoft.Json can deserialize enum from both int value and name; System.Text.Json handles int value only

Can we live with these limitations?

[...]

PR #848, opened 2021-06-02
System.Text.Json support

Yaevh:
System.Text.Json (#678)


rasmus:
Build is broken, I'll have a look

Yaevh:
@rasmus I'ver fixed the tests, can you take a look?

System.Text.Json has two basic limitations compared to Netwonsoft.Json:
1. System.Text.Json treats empty JSON as invalid and throws JsonException; Newtonsoft.Json deserializes it to null
2. Newtonsoft.Json can deserialize enum from both int value and name; System.Text.Json handles int value only

As a result, tests of System.Text.Json do not cover all the aspects covered by Newtonsoft.Json. I don't however think that those two features of Newtonsot.Json are used anywhere by EventFlow.

rasmus:
This is great @Yaevh there's still some evaluation if one of the JSON support should be moved to a separate NuGet package (maybe Newtonsoft), but getting this in will make it a whole log easier.

Yaevh:
Regardless of the outcome of #876, could you please merge this PR? We can always extract the code to a different project afterwards, and I'm avaliable to do this if needed :)

BTW, I believe there are no breaking changes in this PR, so maybe we could add `System.Text.Json` support to V0 already?

rasmus:
I'll have a look

rasmus:
Regardless if this is added or not, adding the `SystemTextJsonSerializer` class to your own project is still possible as this operates on public interfaces.

PR #849, opened 2021-06-07
Add support for using a custom hangfire queue name

dco123:
I've adapted the solution from rikbosch here: https://discuss.hangfire.io/t/one-queue-for-the-whole-farm-and-one-queue-by-server/490/4

The main requirement for allowing this is that we have shared development databases used by many developers. Allowing a developer to use a custom queue name allows them to ensure that they are going to handle the events / jobs they raise, so it can be properly debugged. Without it there's no guarantee where it will run if multiple developers are debugging at the same time.

Usage is:
```csharp
eventFlowOptions.UseHangfireJobScheduler(o => o.UseQueueName("myqueue"))
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,11 @@
-### New in 0.82 (not released yet)
+### New in 0.83 (not released yet)
+
+* New: Queue name used by HangfireJobScheduler can be overridden:
+  ```csharp
+  eventFlowOptions.UseHangfireJobScheduler(o => o.UseQueueName("myqueue"))
+  ```
+
+### New in 0.82.4659 (released 2021-06-17)
 
 * Fix: Source IDs are now added to snapshots
 * Fix: InMemoryReadStore will not break on unmodified update result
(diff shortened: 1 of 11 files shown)


rasmus:
❤️  I'll have a look

rasmus:
Build is broken, I'll have a look

rasmus:
@dco123 there's a few braking tests

dco123:
@rasmus apologies, I didn't have sql server installed so relied on my own testing. I thought the IoC layer would respect optional parameters but it doesn't, so rather than change that i've just registered a default value.

dco123:
I don't think the latest test failure is related to my change. I haven't changed anything outside of EventFlow.Hangfire

dco123:
@rasmus could you take another look when you get chance? I don't think the test failure is related to my change and I notice it is ignored in develop-v1 so probably a known issue.

Thanks

dco123:
@rasmus can you take a look please?

rasmus:
@dco123 I merged `develop-v0` into your PR, which should have the test fixes

rasmus:
Perfect, could you update the release notes we well?

I think this high number arguments that keep getting added to the `IHangfireJobRunner` indicates that it should be refactored for 1.x with some sort of builder pattern.

dco123:
Thanks @rasmus. I've updated the release notes and bumped the version to 0.83 (I think this was needed, let me know if not).

I added an overload to IHangfireJobScheduler to keep backwards compatibility with existing jobs, since they'll be trying to invoke the existing method, but maybe you could have an arguments class instead to avoid having to keep adding overloads. Just need to ensure changes are backwards compatible.

PR #850, opened 2021-06-09
GitHub Actions

rasmus:


rasmus:
Using Linux containers on Windows GitHub Action runners is currently not possible 😢

PR #888, opened 2021-07-11
Proposal: move from .net standard 2.0 to .net standard 2.1 to allow more memory-efficient solutions

madsiberian:
Hello! I think that there is a room for more memory-efficient solutions throughout the EventFlow, but it requires moving to .net standard 2.1 (for proper Span support).
Here is example of how GuidFactories.Deterministic could be optimized.

Benchmark: https://gist.github.com/madsiberian/f31c3b1d26606b4bd39e5bdb100a7124

Before:
<html>
<body>

Method | Mean | Error | StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated
-- | -- | -- | -- | -- | -- | -- | --
Run | 1.441 s | 0.0284 s | 0.0450 s | 90000.0000 | 1000.0000 | - | 854 MB

</body>
</html>
After:
<html>
<body>

Method | Mean | Error | StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated
-- | -- | -- | -- | -- | -- | -- | --
Run | 881.0 ms | 1.55 ms | 1.45 ms | - | - | - | 6 KB

</body>
</html>

Environment:
```
BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19043.1083 (21H1/May2021Update)
Intel Core i9-10900X CPU 3.70GHz, 1 CPU, 20 logical and 10 physical cores
.NET SDK=5.0.301
  [Host]     : .NET 5.0.7 (5.0.721.25508), X64 RyuJIT
  DefaultJob : .NET 5.0.7 (5.0.721.25508), X64 RyuJIT
```

P.S. This is my first open source contribution (ever), so please point out any mistakes -- I am ready to change.


rasmus:
.NET Standard would also make it possible to use `IAsyncEnumerable` 😄

rasmus:
.NET Core 2.1 has reached end of life, thus dropping support for it might be a possibility 
https://dotnet.microsoft.com/platform/support/policy/dotnet-core

PR #890, opened 2021-08-26
New feature: Entity Framework - Eager loading of related data

danm-de:
Added extensions methods to configure eager loading of related data by appending `.Include(..)` / `.ThenInclude(..)` statements.
(see Microsoft documentation https://docs.microsoft.com/en-us/ef/core/querying/related-data/eager.)

The Entity Framework adds FOREIGN-KEY constrains when working with navigation properties. This can lead to ordering problems when saving the read models within a single command. The problem is, that read models are processed separately - each using its own `DbContext` instance. If there is a relation between a _parent_ and its _child_ object, the _parent_ MUST be saved first. This cannot be enforced when a single aggregate update results in more than one read model updates.

The usage is pretty simple:

```c#
 public static IEventFlowOptions Configure(this IEventFlowOptions options)
 {
    return options
        .UseEntityFrameworkReadModel<MyEntity, MyDbContext>(
            cfg => cfg.Include(x => x.SomeProperty).ThenInclude(y => y.SomeOtherProperty)
        );
}
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,18 @@
 
 * Fix: Source IDs are now added to snapshots
 * Fix: InMemoryReadStore will not break on unmodified update result
+* New: added extension methods to the `EventFlow.EntityFramework` package that allow
+  us to configure [eager loading of related data](https://docs.microsoft.com/en-us/ef/core/querying/related-data/eager). Example usage:
+  ```csharp
+  public static IEventFlowOptions Configure(this IEventFlowOptions options)
+  {
+    return options
+      .UseEntityFrameworkReadModel<MyEntity, MyDbContext>(
+        cfg => cfg.Include(x => x.SomeProperty)
+                  .ThenInclude(y => y.SomeOtherProperty)
+      );
+  }
+  ```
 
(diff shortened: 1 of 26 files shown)


danm-de:
``` 
- Coverage       72.47%   39.57%   -32.90%  
``` 

image

rasmus:
The codecov data is a bit outdated as their integration have been broken (which they seem to have fixed now)

rasmus:
Hi @danm-de 

Thanks for the PR, and yes I know, there's a license header in every single file 🤷 legal stuff is hard. Some EventFlow code got copied a while back, and event though its MIT, the license still wasn't present.

Can you update the release notes as well, then I can get it merged and released.

If you are up for it, I'm still lacking the 1.x port of the project. Its a though job, but I haven't worked with Entity Framework on production systems for ages and I might miss some detail.

danm-de:
Thanks. Is the information in the release notes sufficient or is something missing? 

My colleagues and I are currently trying out EventFlow for the first time in a small test project. Please give us some time. We have to understand the concept, the code base etc. correctly before we can adapt / port `EventFlow.EntityFramework` to v1.

rasmus:
❤️

PR #893, opened 2021-08-27
Allow the use of explicitly implemented interfaces in the read model

### LINKED ISSUE #803: ReadModel with explicitly implemented interface is not updated
Hi,

Just getting started with EventFlow but found an issue that tripped me up.

If an `IReadModel` explicitly implements the `IAmReadModelFor<Aggregate, Identity, Event>` interface, the Apply method is for this interface is not called.

For example:

```c#
public class PersonIdentifier : IReadModel, IAmReadModelFor<Aggregate, Identity, Event>
{
    void IAmReadModelFor<Aggregate, Identity, Event>.Apply(IReadModelContext context, IDomainEvent<Aggregate, Identity, Event> domainEvent)
    {
        // this will never be called
       throw new NotImplementedException();
    }
}
```

I believe it may be related to how the method is dynamically invoked from here: https://github.com/eventflow/EventFlow/blob/3c09676a867956158c34f2afff06683f4950df9e/Source/EventFlow/ReadStores/ReadModelDomainEventApplier.cs#L66

Thanks for the awesome project!

danm-de:
If there is no public 'Apply' method (with the expected method parameters), the algorithm will check whether the read model implements either IAmReadModelFor or IAmAsyncReadModelFor. If this is the case, the 'Apply' method of the interface type is called instead.

Fixes #803

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 
 * Fix: Source IDs are now added to snapshots
 * Fix: InMemoryReadStore will not break on unmodified update result
+* Fix: Allow the use of explicitly implemented interfaces in the read model
 
 ### New in 0.81.4483 (released 2020-12-14)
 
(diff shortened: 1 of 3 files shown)


rasmus:
❤️

PR #924, opened 2022-02-11
Fix Cake build

rasmus:


rasmus:
Killed in favor of #925 (for now)

PR #967, opened 2023-01-20
v0: Give life back

rasmus:


rasmus:
Had forgotten that GitHub Actions doesn't support Linux containers on Windows😢

Issue #876, opened 2021-06-17
V1 API: Which should be the default serializer 


rasmus:
With the addition on #848 for v1, there's now an option to switch which is the default. In addition, it might make sense to move e.g. Newtonsoft to a separate NuGet package to minimize the dependencies referenced by the core `EventFlow` package.

SeWaS:
Hello there,

If I may ask a question: would you leave the interface `IJsonSerializer` in the core `EventFlow` package and then offer two NuGet packages, `EventFLow.Newtonsoft` and `EventFlow.SystemTextJson` (for the lack of a better naming), for instance? Or would you go with one package like `EventFlow.Serializing` which provides the public interface, as well as multiple implementations? Or something in between 😃?

I am just curious about the dependency structure you have in mind. Looking forward to the V1 release!!

CentinelaBot:
Hello,

Would having it separate still require us to install it for EventFlow to work properly?

If the answer is yes, isn't it better to leave it as a dependency from EventFlow. You will still end up having to install it anyway, unless you are supporting more than one serialization framework.

Cheers.

kyle-bradley:
System.Text.Json has matured significantly since this discussion and has become the standard in most projects. I would suggest having it as default with Newtonsoft being a separate package.

rasmus:
@kyle-bradley definitely! I think I'll properly have it as a required setting, i.e., none selected by default. Then you have to set one or the other. While inconvenient for new projects, it makes it more specific for developers upgrading from 0.x to 1.x. The error message should then have clear instructions on what to do.

PR #983, opened 2023-04-16
System.Text.Json made as default provider

kyle-bradley:
In reference to Issue #876 I have made System.Text.Json as the default serializer with Newtonsoft moved as its own package.

System.Text.Json is stricter in the way it serializes/deserializes so had to minor adjustments to the way ISourceId is handled in the tests and example code.


rasmus:
Will have a look 👁️

PR #985, opened 2023-05-08
Migrate Event Store & System.Text.Json

kyle-bradley:
I by mistake combined these two changes into one, but they are related so haven't taken the effort in separating them - unless it is requested.

See https://github.com/eventflow/EventFlow/pull/983 For the description on System.Text.Json changes.

I upgraded Event Store and took the opportunity to upgrade most of the other packages. MsSQL and PostgreSQL being the exception as those would be a larger migration.


rasmus:
I'll have a look On 8 May 2023, at 16.28, CLAassistant ***@***.***> wrote:﻿
 Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our Contributor License Agreement before we can accept your contribution.Kyle Bradley seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please add the email address used for this commit to your account.You have signed the CLA already but the status is still pending? Let us recheck it.

kyle-bradley:
Sorry about that, had the wrong email address linked

Source/EventFlow.EventStore/EventFlow.EventStore.csproj
```diff
@@ -21,11 +21,22 @@
...
-    <PackageReference Include="EventStore.Client" Version="5.0.1" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\EventFlow\EventFlow.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
```
joshua211: The PackageReference is the same for` net6.0 `and `net7.0` ?

kyle-bradley: I kept the same condition approach as previous projects, but could be simplified in PackageReference

joshua211 (approved):

Jaben:
@rasmus So, no upgrade to System.Text.Json?

rasmus:
@Jaben it's the one thing that I'm most scare of doing, simply removing the attributes from the values object classes. Doing this can potentially break it for a lot of people and with everything going on with the 1.0 preparation I'm looking to push this to a 2.0 release instead.

However, I will be looking at this PR and another attempt that has been done in order to add `System.Text.Json` support baked into 1.0 at "launch". The only downside of this, is that any releases that use `System.Text.Json` will have a dead `Newtonsoft.Json`  DLL along for the ride. Yes, not optimal, but people that really want `System.Text.Json` get it and people on `Newtonsoft.Json` gets an easier upgrade path.

Again, this is a change I have had a really, really hard figuring out how to make every one happy and merely having the dead DLL seems like the lesser evil for the 1.0 release.

Jaben:
@rasmus Thanks for the response! It does look like the PR(s) have massive surface areas. I understand the concern for backward compatibility and that it's not necessarily worth tackling for 1.0 as it's a feature. But I wouldn't mind the dead "Newtonsoft.Json" dependency option as a half-step.

PR #987, opened 2023-05-13
Event loading as a separate task as processing it onto the read side

kyle-bradley:
In production with 100k+ events, having it pull all the events into memory first before applying them to the read side causes a lengthy processing time as well as high memory usage. 

Also, populate doesn't have built-in "Rebuild" capabilities to process multiple read models at the same time.


rasmus:
@kyle-bradley Awesome! could you update the release notes as well to ensure that developers know about this? ❤️

kyle-bradley:
Will do - I also see that my unit test failed. I did a final load test and forgot to revert. Will fix that in a bit.

Issue #911, opened 2021-11-10
ReadModel Versioning issue.


trikk86:
The problem is probably easy to solve but I spent some time on it and I can't figure it out and EventFlow ReadTheDocs is very outdated.

I created a ReadModel class that implements IAmAsyncReadModel for 3 events and IReadModel interface. Everything goes smooth but I would like to have the ReadModel version.

I added a property to my ReadModel

`        [SqlReadModelVersionColumn]
        public int? Version { get; set; }`

Whenever I create a new row it works but after the row is updated I get an error about declaring the scalar @PREVIOUS_VERSION. I tried making it a field, changing the type to long?/long/int and nothing works.

If I remove the whole column + property it works with no issues

rasmus:
I'll have a look at it this weekend

rasmus:
@trikk86 I tested the read model versioning in a simplified test in PR #915 and its updated as expected. Could you create a similar test in the EventFlow repo that shows your problem?

trikk86:
Hi @rasmus 
I will look into it and do more tests/look into it. I know I had a project in earlier versions where it worked with no issues so I'll try to figure it out tomorrow :)
Thanks for the help!

trikk86:
I looked into this and:

I am using version 0.83.4713 and the problem was that you have to use the explicit MsSql attribute. When you use the SqlReadModelVersionColumn attribute it messes up with the GetVersion of the ReadModel store.

Thank you very much @rasmus  and sorry to have wasted some of your time :(

PR #990, opened 2023-05-23
Update Redis package missing fields

rasmus:
Added all the missing .NET versions... might be too many changes


rasmus:
Doing #911 instead

PR #995, opened 2023-06-05
Migrate RabbitMq to V1

kyle-bradley:
I will also be upgrading the RabbitMq implementation to dispatch both events and commands. Allowing both to be sent over a message queue. I sometimes employed this strategy for sending commands from a Sagas back to the respective services.


rasmus:
The `TreatWarningsAsErrors` is removed, something that is required. Removing it lets all sorts of small errors make their way into the code base.

kyle-bradley:
> The `TreatWarningsAsErrors` is removed, something that is required. Removing it lets all sorts of small errors make their way into the code base.

It was moved in the main RabbitMq project to align with the format of the other projects. Only removed it from the Tests project as I based the build settings on the other extension test projects. 

However, I see that you have updated these settings across the board. Will merge and make sure that these changes are reflected correctly.

Issue #919, opened 2022-01-18
Snapshot upgrader need metadata


nlips:
My SnapshotUpgrader need to retrieve data from the aggregate's events.
For this, it use the IEventStore instance.
But it should only take into account events whose sequence is less than or equal to that of the snapshot.
This information should be an argument of the UpgradeAsync method.
Moreover IEventStore should provide a method to get a range of events.

nlips:
Actually I can't use IEventStore... I don't have the aggregate's id !

SeWaS:
Hello there 👋 

Long-time lurker here. Decided that it's time to contribute a bit to the repository.

The underlying idea of enabling the event store to load domain events within a given sequence-number range makes sense to me. I would go ahead and prepare a PR if still relevant.

Regarding your snapshot upgrading question, would it help if the specific snapshots in your application would carry their respective aggregate Id in the payload? That way you would have access to it when upgrading from one snapshot version to the other, no? 
In case you found a solution to the problem, I would be curious what it was :) 

Cheers

rasmus:
While an old issue, having more metadata available in the snapshot upgraders makes a lot of sense. Since it involves breaking changes, its a good candidate for a v2.

rasmus:
@nlips while it doesn't save you from your current problem, you could add the information you need to the snapshot yourself, e.g. add a property that holds the aggregate ID or other meta data you might need.

PR #1008, opened 2023-10-30
Issue #919: Enable IEventStore to load events up to a given sequence number

SeWaS:
Hello 👋 

As mentioned in issue #919 , let's enhance the event store API to be able to load events which are within a given sequence-number range of the targeted aggregate.

If this PR could help, just let me know when things are missing, docs have to be adjusted or if any more tests are required. Would be happy to adjust.

Cheers


kyle-bradley:
Great feature! Especially when it comes to debugging. 

One suggestion would be to extend this functionality to the rebuilder. That way, you can get your write & read state to a certain point in time.

PR #1037, opened 2024-07-25
v0: Added overload of `MongoDbOptionsExtensions.ConfigureMongoDb` that passes IResolverContext through to IMongoDatabase factory delegate

dco123:
We need access to the service provider in order to resolve the IMongoDatabase, seems like exposing the resolver would be the best way to do that


rasmus:
I'm working on getting a GHA runner up and running for EventFlow that has Docker properly configured (Linux containers).

rasmus:
Builds up and running again @dco123. Will hopefully be able to merge and release this soon

PR #1062, opened 2024-12-03
v1: Provide event naming strategies

### LINKED ISSUE #907: Fix: Supply another mechanism for event naming
As mentioned in #903, events with the same name created in different namespaces will collide. While there is a workaround using `[EventVersion]`, the OOB experience should be that it should merely work.

An idea would be to create some settings that developers are forced to do during EventFlow initialization, like selecting the naming for events. Then introduce a new naming convention named `NamespaceAndClassName` along side the existing `ClassName` naming.

SeWaS:
Related to #907, might even close #907.

This PR provides the capabilities to EventFlow to apply naming strategies which can ensure that no collisions among events with the same class name but different namespaces occur.

There are a few built-in strategies, users can provide their own strategy by simply implementing the `IEventNamingStrategy` interface.

The desired naming strategy needs to be registered to the DI container, otherwise a `ViodStrategy` is registered as default, which represents the current naming strategy.

Also, I added a simple and small README for the documentation.


Source/EventFlow/Configuration/IEventFlowConfiguration.cs
```diff
@@ -86,5 +87,17 @@ public interface IEventFlowConfiguration
...
+        /// Use <c>EventNamingStrategy</c> to determine the internal naming strategy
+        /// for events. The strategy is applied whenever an event is persisted
+        /// and loading from persistence and when serializing/deserializing in that
+        /// context respectively.
+        /// 
+        /// If more fine grained control of is needed, a custom implementation of
+        /// <c>IEventNamingStrategy</c> should be provided.
+        /// </summary>
+        /// <remarks>Defaults to SimpleNamingStrategy</remarks>
+        IEventNamingStrategy EventNamingStrategy { get; }
```
rasmus:
```suggestion
```

Don't think this is used anywhere, might be leftovers from the original commit

Source/EventFlow/Configuration/EventFlowConfiguration.cs
```diff
@@ -86,6 +87,8 @@ public class EventFlowConfiguration : IEventFlowConfiguration, ICancellationConf
         public CancellationBoundary CancellationBoundary { get; set; }
         
         public bool ForwardOptimisticConcurrencyExceptions { get; set; }
+        
+        public IEventNamingStrategy EventNamingStrategy { get; set; }
```
rasmus:
```suggestion
```

Same as the implementation, don't think this is used.

Source/EventFlow/Configuration/EventNamingStrategy/VoidStrategy.cs
```diff
@@ -0,0 +1,31 @@
...
+// FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+
+namespace EventFlow.Configuration.EventNamingStrategy
+{
+    internal class VoidStrategy : IEventNamingStrategy
```
rasmus: `DefaultStrategy` might be a better name. For me at least, the word "void" refers to something being empty, or non-existing. Thus there might be an expectation that it simply would name the events anything and should be replaced always.

Source/EventFlow/Configuration/EventNamingStrategy/NamespaceAndNameStrategy.cs
```diff
@@ -0,0 +1,32 @@
...
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+
+namespace EventFlow.Configuration.EventNamingStrategy
+{
+    public class NamespaceAndNameStrategy : IEventNamingStrategy
+    {
+        public string CreateEventName(int version, Type eventType, string name) 
+            => eventType.Namespace + "." + name;
```
rasmus:
```suggestion
            => $"{eventType.Namespace}.{name}";
```
I seem to recall that you should the `$` or `string.Format` as it has less memory allocations. I can't find the reference though.

Source/EventFlow/Configuration/EventNamingStrategy/NamespaceAndClassNameStrategy.cs
```diff
@@ -0,0 +1,32 @@
...
+// CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+
+namespace EventFlow.Configuration.EventNamingStrategy
+{
+    public class NamespaceAndClassNameStrategy : IEventNamingStrategy
+    {
+        public string CreateEventName(int version, Type eventType, string name) =>
+            eventType.Namespace + "." + eventType.Name;
```
rasmus:
```suggestion
            $"{eventType.Namespace}.{eventType.Name}";
```

I seem to recall that you should the `$` or `string.Format` as it has less memory allocations. I can't find the reference though.

rasmus (changes requested):
Awesome! Many thanks for the PR, just some minor comments and nitpicks.

Source/EventFlow/Configuration/IEventFlowConfiguration.cs
SeWaS: Right, simply leftovers. I had them actually on my "to do before push" list

rasmus (approved):

rasmus:
❤️

rasmus:
I'll get this added, update release notes and released

PR #1069, opened 2025-01-05
Documentation improvements: Aggregates

alexeyfv:
Hello,

I made the following improvements to the `Aggregate` page:

1. Added a link to the `Identity` page. Seems like it was missing.
2. Briefly explained why it's important to use the `SingleValueObjectConverter` class with the `Identity` class.
3. Changed the code snippet in the "Using Events" section to make it more clear.
4. Added examples of how to modify aggregates using the IAggregateStore and the CQRS approach.


rasmus:
Some very odd comments from copilot. Testing it out to see if it actually has any value.

rasmus:
Thanks for the updates. It really means a lot!

Issue #1074, opened 2025-01-08
Asynchronous jobs don't work as expected


alexeyfv:
Hi,

I think I found two potential issues in the jobs functionality. But maybe I don’t understand something. Please let me know if I’m wrong.

### Code snippets

There is a test aggregate.

```csharp
class Test(TestId id) : AggregateRoot<Test, TestId>(id), IEmit<Pinged>
{
    public Task Ping()
    {
        Console.WriteLine("Ping");
        Emit(new Pinged());
        return Task.CompletedTask;
    }

    public void Apply(Pinged aggregateEvent) { }
}
```

There is also a subscriber to `Pinged` event. The subscriber schedules a job with 5 sec delay.

```csharp
class TestSubscribers(IJobScheduler jobScheduler) : ISubscribeAsynchronousTo<Test, TestId, Pinged>
{
    public async Task HandleAsync(IDomainEvent<Test, TestId, Pinged> domainEvent, CancellationToken cancellationToken)
    {
        var job = new PingJob() { TestId = domainEvent.AggregateIdentity };

        Console.WriteLine("Schedule job");

        await jobScheduler.ScheduleAsync(job, TimeSpan.FromSeconds(5), cancellationToken);
    }
}

```

The job just pings the aggregate. And the whole process repeats.

```csharp
class PingJob : IJob
{
    public required TestId TestId { get; init; }

    public async Task ExecuteAsync(IServiceProvider serviceProvider, CancellationToken cancellationToken)
    {
        Console.WriteLine("Job started at: {0}", DateTimeOffset.Now);

        await serviceProvider
            .GetRequiredService<IAggregateStore>()
            .UpdateAsync<Test, TestId>(TestId, TestId.New, (a, _) => a.Ping(), cancellationToken);
    }
}
```

Full example is [here](https://github.com/alexeyfv/EventFlowJobsExample).

### 1st issue

From the code above, I expect that the job will be started after 5 sec delay. But in my case, it starts only after 15 sec.

```sh
dotnet run
Ping
Hangfire Server started. Press ENTER to exit...
Schedule job
Job started at: 08.01.2025 19:18:42 +01:00
Ping
Schedule job
Job started at: 08.01.2025 19:18:57 +01:00
Ping
Schedule job
Job started at: 08.01.2025 19:19:12 +01:00
Ping
Schedule job
Job started at: 08.01.2025 19:19:27 +01:00
Ping
Schedule job
```

### 2nd issue

[...]

PR #1093, opened 2025-03-27
Fix multiple calls of the same async subscribers when dispatching events

alexeyfv:
Hi,

I found the issue with async subscribers (#1074). When using the `AddDefaults` method, it registers the same subscriber multiple times. During event dispatching, duplicates were not removed, causing the same subscriber to be invoked multiple times.

Let me know if you have any feedback.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 ### New in 1.2.1 (working version, not released yet)
 
-* *Nothing yet...*
+* Fix: Prevent multiple calls of the same async subscribers when dispatching events (by @alexeyfv)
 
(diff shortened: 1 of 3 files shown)


rasmus:
@alexeyfv can you update the release notes as well? I'll get this merged. Sorry, private life has hit hard and has removed a lot of focus from this project (again). Will try to catch up

alexeyfv:
The pipeline failed. Need to [update](https://github.com/actions/runner-images/issues/11101) the image version in [pipleine.yaml](https://github.com/eventflow/EventFlow/blob/e3c50c58d755439f67a87b6adb486348c34d6fb3/.github/workflows/pipeline.yaml#L30).

PR #1098, opened 2025-05-15
Add domain event publisher to AggregateStore

AlexGreatDev:
- Introduced a new private field `_domainEventPublisher` in `AggregateStore` and updated the constructor to use it directly for publishing domain events, removing the previous service provider retrieval.
- Refactored the `CanConvert` method in `SingleValueObjectConverter` to use expression-bodied syntax for improved readability while maintaining its original functionality.

### DIFF
--- a/Source/EventFlow/Aggregates/AggregateStore.cs
+++ b/Source/EventFlow/Aggregates/AggregateStore.cs
@@ -52,6 +52,7 @@ namespace EventFlow.Aggregates
         private readonly ICancellationConfiguration _cancellationConfiguration;
         private readonly IAggregateStoreResilienceStrategy _aggregateStoreResilienceStrategy;
         private readonly IEventFlowConfiguration _eventFlowConfiguration;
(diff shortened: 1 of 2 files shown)


rasmus:
Its not possible to add this dependency as it will cause a circular dependency

```
System.InvalidOperationException : A circular dependency was detected for the service of type 'EventFlow.Aggregates.IAggregateStore'.
[08:38:31 INF] EventFlow.Aggregates.IAggregateStore(EventFlow.Aggregates.AggregateStore) -> EventFlow.Subscribers.IDomainEventPublisher(EventFlow.Subscribers.DomainEventPublisher) -> EventFlow.Sagas.IDispatchToSagas(EventFlow.Sagas.DispatchToSagas) -> EventFlow.Sagas.ISagaStore(EventFlow.Sagas.AggregateSagas.SagaAggregateStore) -> EventFlow.Aggregates.IAggregateStore
```

PR #1108, opened 2025-06-14
Replace FluentAssertions with Shouldly (#1081)

### LINKED ISSUE #1081: Replace FluentAssertions with Shouldly
Open-source should remain open-source and as a MIT licensed project, EventFlow very much beleives in being **open** and free for everyone!

While there certainly is a problem regarding companies not providing any support (monetary nor PRs) to the open-source projects they use, what FluentAssertions has done is *not* the way. That simply causes too much uncertainty and too much work for everyone.

- https://github.com/fluentassertions/fluentassertions/pull/2943 
- https://xceed.com/products/unit-testing/fluent-assertions/

Image

Going forward, EventFlow would like to show its support via usage and downloads to the [shouldly](https://docs.shouldly.org/) library instead.

So, you are up for it. Pick some tests and convert them to shouldly, it shouldn't be all or even all within a single project. As it is a rather big undertaking, I expect that both assertion libraries will exist in the codebase for some time.

---

As a side note, if you are a company and use EventFlow (or other open-source), really consider showing your support. Specifically, EventFlow has only received very few PRs from companies and no monetary support.

Focus1337:
Closes #1081 

Replaced FluentAssertions with Shouldly


rasmus:
❤️ will have a look

PR #1112, opened 2025-06-19
Cleanup dependencies #1107

### LINKED ISSUE #1107: Leverage dependencies available from the framework
in EventFlow

### Is your feature request related to a problem?

I want to minimise dependencies in my project by utilising framework dependencies wherever possible

### What solution would you like?

I want Microsoft.CSharp to no longer be an explicit dependency on net 6+ etc

### Describe alternatives you've considered

Accept the additional dependency

### Additional context

Have only checked EventFlow but all packages should be checked to see what optimisations are possible

thompson-tomo:
Closes #1107


rasmus:
Suspicious PR

rasmus:
😅 re-investigating this

PR #1113, opened 2025-06-19
Fix solution file

thompson-tomo:


rasmus:
Sorry. Only items that are marked as _done_ and ready are included in the solutions file to make it easier to on things that are working.

See https://github.com/eventflow/EventFlow?tab=readme-ov-file#nuget-package-status

